PR #3, opened 2017-12-21
APF-680 Reorganize packages for Javadoc

rworsnop:


common/src/main/java/com/vmware/connectors/common/web/SingleDeferredResult.java
```diff
@@ -3,8 +3,9 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-package com.vmware.connectors.common.utils;
+package com.vmware.connectors.common.web;
 
+import com.vmware.connectors.common.web.DeferredResultSubscriber;
```
RyanBard: Unnecessary import?

common/src/main/java/com/vmware/connectors/common/web/SingleReturnValueHandler.java
```diff
@@ -3,8 +3,9 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-package com.vmware.connectors.common.utils;
+package com.vmware.connectors.common.web;
 
+import com.vmware.connectors.common.web.SingleDeferredResult;
```
RyanBard: Here too? (unnecessary)

connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
```diff
@@ -7,7 +7,7 @@
 
 package com.vmware.connectors.concur;
```
RyanBard: Could you fix the copyright file header above while you're re-org'ing.  Maybe do a search through the code in case we missed some other places too.

RyanBard:
If you don't get to it tonight (I'm assuming you're done for the night), I'll merge this and do it in a followup on Wednesday.

common/src/main/java/com/vmware/connectors/common/web/SingleDeferredResult.java
rworsnop: I guess PMD needs fixing in this project as well.

RyanBard: It turns out that pmd isn't running as part of the build.  It's due to something really silly, however, after turning it on, I'm finding lots of places that need code cleanup, so I won't do it in this PR.

I'll merge this and create a ticket to fix the issues.

connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
RyanBard: I'll get this in the followup ticket/PR.

PR #14, opened 2018-01-09
Tidy up READMEs

rworsnop:


README.md
```diff
@@ -29,10 +30,57 @@ Additionally, it is possible to build RPMs, which install the connectors as serv
...
+echo "security.oauth2.resource.jwt.key-uri=https://acme.vmwareidentity.com/SAAS/API/1.0/REST/auth/token?attribute=publicKey&format=pem" \
+> /etc/opt/vmware/connectors/jira/application.properties
+```
+The hostname of the URL will vary depending on your IDM tenant.
+
+The configuration file created above must be part of the `roswell` user and group. Again, using Jira as an example:
+```
+chown roswell:roswell /etc/opt/vmware/connectors/jira/application.properties
+```
+The connector being a Spring Boot application, many other configuration options are available&mdash;for example, `server.port`. Please see the Spring Boot documentation for more details.
```
RyanBard: Is `&mdash;` leftoever from an xml or html or something?

README.md
```diff
@@ -29,10 +30,57 @@ Additionally, it is possible to build RPMs, which install the connectors as serv
...
+
+There might also be connector-specifc configuration required. Please see the README files within the [individual connectors](https://github.com/vmware/connectors-workspace-one/tree/master/connectors) for further details.
```
RyanBard: Typo: specifc -> specific

connectors/servicenow/README.md
```diff
@@ -1,86 +1,7 @@
 # ServiceNow Connector
 
-## Building
+The ServiceNow connector presents cards inviting the user to approve or reject open requests. It does this based on request numbers passed as tokens in the card request.
```
RyanBard: I forgot to mention that the user's email is also passed in: https://github.com/vmware/connectors-workspace-one/blob/master/connectors/servicenow/src/main/resources/static/discovery/metadata.hal#L6

Please change this from `... based on request numbers passed as tokens in the card request` to `... based on request numbers passed as tokens and the user's email address in the card request`.

README.md
rworsnop: It's the easiest way I know of to display an em dash.

RyanBard: Ah, I didn't realize that was markdown (it looked like xml/html).

rworsnop: It is HTML, as there's currently no way to do it with markdown. I guess there's some kind of pass-through capability at play here.

PR #16, opened 2018-01-12
APF-711 Add check for RPM

rworsnop:
This change adds a check to ensure that the connectors can be installed
using the generated RPMs.

As part of the build (when the check-rpm profile is active) we now spin
up a Centos 7 Docker container and install and configure the connector.
If the connector's health check doesn't succeed within 60s, the build will fail.

### DIFF
--- a/connectors/airwatch/src/test/resources/application.properties
+++ b/connectors/airwatch/src/test/resources/application.properties
@@ -1,1 +1,0 @@
-security.oauth2.resource.jwt.key-uri=https://acme.vmwareidentity.com/SAAS/API/1.0/REST/auth/token?attribute=publicKey&format=pem
--- a/pom.xml
+++ b/pom.xml
@@ -451,6 +451,71 @@
                         <autoReleaseAfterClose>true</autoReleaseAfterClose>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>io.fabric8</groupId>
+                    <artifactId>docker-maven-plugin</artifactId>
+                    <version>0.23.0</version>
+                    <configuration>
+                        <images>
+                            <image>
+                                <name>connector</name>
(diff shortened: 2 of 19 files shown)


connectors/airwatch/src/test/resources/application.properties
```diff
@@ -1 +0,0 @@
-security.oauth2.resource.jwt.key-uri=https://acme.vmwareidentity.com/SAAS/API/1.0/REST/auth/token?attribute=publicKey&format=pem
```
RyanBard: I seem to remember needing this for the tests to pass.  Was this just memory of the old docker container tests? (The unit tests aren't hitting whatever code that asks for vIdmPubKey and are still passing?)

pom.xml
```diff
@@ -451,6 +451,71 @@
...
+                                    <cmd>
+                                        <shell>/usr/sbin/init</shell>
+                                    </cmd>
+                                </build>
+                                <run>
+                                    <privileged>true</privileged>
+                                    <ports>connector.port:8080</ports>
+                                    <wait>
+                                        <!-- Build will fail if not healthy within 60 seconds -->
+                                        <time>60000</time>
```
RyanBard: In practice, about how long does this take to check the rpm?

connectors/airwatch/src/test/resources/application.properties
RyanBard: Ah I see.  Not shown in this diff is the key-value is set in the connectors-test's properties file: https://github.com/vmware/connectors-workspace-one/blob/master/common/connectors-test/src/main/resources/application-test.properties

pom.xml
rworsnop: It seems to take < 20 seconds.

PR #20, opened 2018-01-18
APF-715: Fix NcssCount in SalesforceController

RyanBard:
* Reorgs the REST calls to be more selective at what we call out to (throughput
  performance via making less wasted calls instead of possible latency
  reduction due to parallel calls for things we won't need)
* Extracts SFAccount and SFOpportunity into their own files
* Makes "private static final" vs. "private final static" consistent
* Reformats constructor, method, and annotation parameters
* Breaks methods down into smaller methods
* Reduces some REST callout and card generation duplication
* Removes unnecessary annotation param names (single default param)
* Removes unnecessary finals (local variables and params in small methods)
* Reorders methods to flow more naturally when reading the code
* Changes code to use a more fluent style
* Changes SFAccount to be immutable

### DIFF
--- a/connectors/salesforce/src/main/java/com/vmware/connectors/salesforce/SalesforceController.java
+++ b/connectors/salesforce/src/main/java/com/vmware/connectors/salesforce/SalesforceController.java
@@ -5,6 +5,8 @@
 
 package com.vmware.connectors.salesforce;
 
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.jayway.jsonpath.DocumentContext;
 import com.jayway.jsonpath.JsonPath;
 import com.vmware.connectors.common.json.JsonDocument;
@@ -16,26 +18,24 @@ import com.vmware.connectors.common.payloads.response.*;
 import com.vmware.connectors.common.utils.Async;
 import com.vmware.connectors.common.utils.CardTextAccessor;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.tuple.Pair;
-import org.pojomatic.Pojomatic;
-import org.pojomatic.annotations.AutoProperty;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.*;
 import org.springframework.util.Base64Utils;
-import org.springframework.util.concurrent.ListenableFuture;
+import org.springframework.util.CollectionUtils;
 import org.springframework.web.bind.annotation.*;
 import org.springframework.web.client.AsyncRestOperations;
-import org.springframework.web.util.UriComponentsBuilder;
 import rx.Observable;
 import rx.Single;
 
 import javax.validation.Valid;
 import java.io.IOException;
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
 import java.util.*;
-import java.util.function.Supplier;
+import java.util.stream.Collectors;
 
 import static org.springframework.http.HttpHeaders.AUTHORIZATION;
 import static org.springframework.http.HttpHeaders.CONTENT_TYPE;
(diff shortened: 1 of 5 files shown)


connectors/salesforce/src/main/java/com/vmware/connectors/salesforce/SalesforceController.java
```diff
@@ -45,16 +45,16 @@
...
+    private static final String SALESFORCE_AUTH_HEADER = "x-salesforce-authorization";
+    private static final String SALESFORCE_BASE_URL_HEADER = "x-salesforce-base-url";
+    private static final String ROUTING_PREFIX = "x-routing-prefix";
     private static final String ADD_CONVERSATIONS_PATH = "/conversations";
     private static final String CONVERSATION_TYPE = "email";
 
-    private final static String WHITE_SPACE = " ";
-    private final static String NEW_LINE = "\n";
-    private final static int SIZE = 256;
+    // TODO: concatenating strings into a SOQL query like this may provide an avenue for a SOQL-injection attack
```
RyanBard: This comment used to be down by the 3 parallel calls (when they use String.format to construct the SOQL).  Now that I moved those around to different places, this made more sense to be near the constants that defined the format string.

connectors/salesforce/src/main/java/com/vmware/connectors/salesforce/SalesforceController.java
```diff
@@ -111,252 +117,281 @@ public SalesforceController(AsyncRestOperations rest, CardTextAccessor cardTextA
         this.sfAttachmentTasklinkPath = sfAttachmentTasklinkPath;
     }
 
-    @PostMapping(path = "/cards/requests", consumes = APPLICATION_JSON_VALUE, produces = APPLICATION_JSON_VALUE)
+    ///////////////////////////////////////////////////////////////////
+    // Methods common to both the cards request and the actions
+    ///////////////////////////////////////////////////////////////////
+
+    private HttpHeaders makeHeaders(String auth) {
```
RyanBard: Most of the methods in this file flow top-down (if A calls into B and C, and B calls into D, and D calls into E, then the order of my functions in source would be: A, B, D, E, C -- depth first traversal).  These are the few methods that are used in more than 1 place (and are relatively small and few in number), so I put them at the top instead.

connectors/salesforce/src/main/java/com/vmware/connectors/salesforce/SalesforceController.java
```diff
@@ -111,252 +117,281 @@ public SalesforceController(AsyncRestOperations rest, CardTextAccessor cardTextA
...
         // TODO: implement a better system of validating domain names than "yup, it's not empty"
-        if (StringUtils.isBlank(senderDomain) || StringUtils.isBlank(userEmail)) {
-            logger.warn("Either sender email or user email is blank for url: {}", sfBaseUrl);
+        if (StringUtils.isBlank(senderDomain) || StringUtils.isBlank(user)) {
+            logger.warn("Either sender email or user email is blank for url: {}", baseUrl);
             return Single.just(new ResponseEntity<>(BAD_REQUEST));
         }
 
-        // TODO: concatenating strings into a SOQL query like this may provide an avenue for a SOQL-injection attack
-        String contactDetailsSoql = String.format(QUERY_FMT_CONTACT, senderEmail, userEmail);
```
RyanBard: These are now being done serial instead of parallel.

The 3 potential REST calls are:
(1) Lookup single contact for combination of user+sender emails
(2) Lookup all contact opportunities for combination of user+sender emails
(3) Lookup all opportunities for the sender's account (because the sender might not exist)

In the previous code: First, it call call out to (1), (2), and (3) and then wait for (1) to return.  If (1) finds a contact, then we throw away the result of (3) and use (2) to create a card with the contact info.  If (1) does not find a contact, then we throw away the result of (2) and use (3) to create a card with an add-contact button.

In this PR, I've changed it to instead: Call out to (1) only then wait.  If (1) finds a contact, call into (2) and return the contact info card.  If (1) does not find a contact, then call into (3) and return the add-contact card.

(1) looks like it would probably be the fastest of the calls (pulling in less information than the others), so I'm choosing potentially a little more latency for better throughput (benefit from making 1/3 less calls).  There was no noticeable difference in my testing, so if there is a difference in latency, it's low milliseconds and/or under the normal variance of calls.

connectors/salesforce/src/main/java/com/vmware/connectors/salesforce/SalesforceController.java
```diff
@@ -111,252 +117,281 @@ public SalesforceController(AsyncRestOperations rest, CardTextAccessor cardTextA
...
-                        HttpMethod.GET, new HttpEntity<String>(headers), JsonDocument.class);
-
-        return Async.toSingle(contactsFuture)
+        return retrieveContactInfos(auth, baseUrl, user, sender)
                 .map(HttpEntity::getBody)
-                .flatMap(contactDetails -> getCards(contactDetails, senderEmail, sfBaseUrl, routingPrefix, headers,
-                        opportunitiesFuture, accountsFuture))
+                .flatMap(contacts -> getCards(contacts, sender, baseUrl, routingPrefix, headers, user, senderDomain))
+                .map(cards -> {
+                    Cards c = new Cards();
```
rworsnop: Please move the braced code into a separate method.

connectors/salesforce/src/main/java/com/vmware/connectors/salesforce/SalesforceController.java
RyanBard: 600ms on the first call (after a restarted jvm, so there is presumably some priming for the connection pool)

sub-100ms on all followup calls (50-80ms or so)

PR #21, opened 2018-01-26
APF 723 implement stash connector

SudharsanThumattiSathyamoorthy:


connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
```diff
@@ -0,0 +1,444 @@
...
+                                                  @RequestHeader(BASE_URL_HEADER) final String baseUrl) {
+
+        logger.debug("Comment action for stash pull request: {}, baseURL: {}",
+                pullRequest,
+                baseUrl);
+
+        final HttpHeaders headers = buildHeaders(authHeader);
+        headers.add(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE);
+
+        final String url = String.format(StashConstants.STASH_PR_COMMENT_URL_FORMAT,
```
RyanBard: It would be better to use spring's UriComponentsBuilder instead of String.format.  Spring will escape/encode the url for us (reducing risks of malicious input causing us to call out to something unexpected).

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
```diff
@@ -0,0 +1,444 @@
...
+        final ListenableFuture<ResponseEntity<JsonDocument>> response = this.rest.exchange(
+                url,
+                HttpMethod.GET,
+                new HttpEntity<>(headers),
+                JsonDocument.class);
+
+        final JsonDocument[] jsonDocuments = new JsonDocument[1];
+        Async.toSingle(response)
+                .subscribe(entity -> {
+                    jsonDocuments[0] = entity.getBody();
```
RyanBard: I'm confused by this.  Isn't this asynchronous and will fall through to the Assert before jsonDocuments[0] is set?

If it happens to be retrieving the comments before falling through, we probably shouldn't be handling this how it's currently written.

We should instead make this in the same asynchronous style as elsewhere.  I'll give a more concrete suggestion after I finish reviewing the rest of the code.

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
```diff
@@ -0,0 +1,444 @@
...
+        final HttpHeaders headers = buildHeaders(authHeader);
+
+        // Get current version of the pull request. Pull request "version" changes when we do any actions on it.
+        // When the pull request is raised, the current value will be 0.
+        // For example, when we approve the pull request, then the version will change from 0 to 1.
+        // We have to add the latest version of the pull request URI to do any actions. Otherwise, the action will be rejected.
+        // If the build for the branch is going on, then the actions would be rejected.
+        final String version = getVersion(headers, baseUrl, pullRequest);
+
+        final String url = String.format(StashConstants.STASH_PR_ACTION_FORMAT,
```
RyanBard: Again, UriComponentsBuilder is safer than String.format.

rworsnop:
Stash has been renamed to Bitbucket Server. It's a bit of a mouthful, but we should get the brand name right. (Note that the "other" Bitbucket is called "Bitbucket Cloud", so we do need the word "server" to differentiate.)

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    private String getVersion(final HttpHeaders headers,
+                              final String baseUrl,
+                              final StashPullRequest pullRequest) {
+
+        final ListenableFuture<ResponseEntity<JsonDocument>> response = getPullRequestInfo(headers, pullRequest, baseUrl);
+        final JsonDocument[] jsonDocuments = new JsonDocument[1];
+        Async.toSingle(response)
+                .subscribe(entity -> {
```
RyanBard: Again, I suspect this won't work.  My reading of the code would expect this to asynchronously fall through and hit the assert before subscribe is called (at least on a real system -- unit test frameworks/libraries/etc. tend to make asynchronous code behave synchronously).

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
```diff
@@ -0,0 +1,444 @@
...
+        return Async.toSingle(stashResponse)
+                .toObservable()
+                .onErrorResumeNext(ObservableUtil::skip404)
+                .map(entity -> convertResponseIntoCard(entity, pullRequest, routingPrefix, comments));
+    }
+
+    private ListenableFuture<ResponseEntity<JsonDocument>> getPullRequestInfo(final HttpHeaders headers,
+                                                                              final StashPullRequest pullRequest,
+                                                                              final String baseUrl) {
+        final String url = String.format(StashConstants.STASH_PR_URL_FORMAT,
```
RyanBard: Another String.format to convert to UriComponentsBuilder.

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
```diff
@@ -0,0 +1,444 @@
...
+                                                              final StashPullRequest pullRequest,
+                                                              final StashAction stashAction) {
+        final HttpHeaders headers = buildHeaders(authHeader);
+
+        // Get current version of the pull request. Pull request "version" changes when we do any actions on it.
+        // When the pull request is raised, the current value will be 0.
+        // For example, when we approve the pull request, then the version will change from 0 to 1.
+        // We have to add the latest version of the pull request URI to do any actions. Otherwise, the action will be rejected.
+        // If the build for the branch is going on, then the actions would be rejected.
+        final String version = getVersion(headers, baseUrl, pullRequest);
```
RyanBard: Do you know if this is for all actions or just some?

In github it was only for merge (optional), in gitlab it was for approve and merge (optional for both).

I remember bitbucket being really obnoxious about some operations in its UI when someone changed something while you were reviewing, but not others.

Also, in github & gitlab, I chose to look up the version when making the cards and to pass that version (commit hash actually) via request-params in the cards' action.  This would cause clicking a button on an old card to error instead of merge something that has changed.

We probably want to do the same for this (or maybe don't specify the hashes in github/gitlab).

It's not clear to me how often the client will ask for an updated card, so maybe the details of this are a bigger discussion for a later time (and including @davidpshaw @jonathanrfisher1).

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
```diff
@@ -0,0 +1,444 @@
...
+                    )
+                    .setType(HttpMethod.POST)
+                    .build()
+                );
+    }
+
+    private String buildActionUrl(final String routingPrefix,
+                                  final StashPullRequest pullRequest,
+                                  final StashAction stashAction) {
+        return String.format(StashConstants.STASH_CLIENT_ACTION_URL,
```
RyanBard: Safer to use UriComponents, but by this point we've probably verified that the pullRequest info is ok (buildActionUrl is called after looking up the pull request and finding it), so not as much of a problem here.

connectors/stash/src/main/resources/static/discovery/metadata.hal
```diff
@@ -0,0 +1,9 @@
+{
+  "authorization_header": "x-stash-authorization",
+  "backend_base_url_header": "x-stash-base-url",
+  "fields": {
+    "pr_email_subject": {
+      "regex": "(([a-zA-Z0-9]+)\/([a-zA-Z0-9-]+) - Pull request #([0-9]+): [a-zA-Z0-9- .\\()#$&@+=_|!]*)"
```
RyanBard: We can reduce this.  We only care about the first 3 pieces, so I think you can get away with:

```
"regex": "(([a-zA-Z0-9]+)\/([a-zA-Z0-9-]+) - Pull request #([0-9]+):[ ])"
```

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
```diff
@@ -0,0 +1,444 @@
...
+                                messageKeyPrefix + ".description",
+                                descriptionArgs
+                        )
+                )
+                .build();
+    }
+
+    private Set<StashPullRequest> convertToStashPullRequest(final Set<String> cardTokens) {
+        final Set<StashPullRequest> pullRequests = new HashSet<>();
+        final Pattern pattern = Pattern.compile(StashConstants.STASH_PR_EMAIL_SUBJECT_REGEX);
```
RyanBard: It's unfortunate that we have to do this, but I don't have a better solution in mind.  We could try to look for the "View pull request" link in the email, but that gets ugly fast:

```
<a href=3D"https://stash.air-watch.com/projects/UFO/repos/app-platform-server=
/pull-requests/252" target=3D"_blank" style=3D"color: #3b73af; text-decoration: =
none; border: 0; text-decoration: none; color: #ffffff">View pull request</a=
>
```
What you have is probably our best option right now.

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
rworsnop: What Ryan said. You almost never want to call `subscribe`.

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
```diff
@@ -0,0 +1,444 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import static com.vmware.connectors.stash.utils.StashConstants.ATLASSIAN_TOKEN;
+import static com.vmware.connectors.stash.utils.StashConstants.COMMENT_PARAM_KEY;
+import static org.springframework.http.HttpHeaders.AUTHORIZATION;
+
+@RestController
+public class StashController {
```
RyanBard: Since Bitbucket is using Basic auth, you'll want to implement a testAuth method for it like Jira:

* Example implementation of test-auth method: https://github.com/vmware/connectors-workspace-one/blob/master/connectors/jira/src/main/java/com/vmware/connectors/jira/JiraController.java#L129
* Example configuration in yml: https://github.com/vmware/connectors-workspace-one/blob/master/connectors/jira/src/main/resources/application.yml#L8

RyanBard: This allows the client to call into the server and verify the credentials that the user typed in were correct immediately rather than at a later point when a card is fetched.

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
RyanBard: Confirmed that this pattern isn't working:

```json
{
  "results": [
    {
      "connector_id": "c5f0d317-9770-42c2-bf53-4102145a5ea9",
      "connector_status": {
        "status": 500,
        "connector_message": "{\"timestamp\":\"2018-01-26T21:36:11.284+0000\",\"status\":500,\"error\":\"Internal Server Error\",\"exception\":\"java.lang.IllegalArgumentException\",\"message\":\"JsonDocument should not be null.\",\"path\":\"/cards/requests\"}"
      }
    }
  ]
}
```

To test, create a connector for this with:

```
curl -v \
     -X POST \
     -H 'Content-Type: application/json' \
     -H "Authorization: Bearer $ADMIN_JWT_TOKEN" \
     -d '
{
    "name": "Stash",
    "external_connector_discovery_url": "http://172.17.100.100:8069",
    "backend_base_url": "https://stash.air-watch.com",
    "authentication": {
        "authentication_type": "BASIC"
    }
}
' \
     "http://$HOST:$PORT/tenantconfig/tenant-connectors"
```

On my box, api-gateway is running on localhost:8080, so $HOST is localhost, $PORT is 8080.  For my external discovery url, I have a vm on 172.17.100.100 (you can create one too with: vagrant up), where I cloned the connectors repo, build with the make-rpm and install that rpm (my application.properties sets the port to 8069).  You can simplify this on your box if you just run the connector with spring boot:run or something.

Whatever ip+port you run the connector on, put in the curl's external_connector_discovery_url.

With that configured, I ran the following cards request to get the above error:

```
curl -v \
     -X POST \
     -H "Content-Type: application/json" \
     -H "Accept: application/json" \
     -H "Authorization: Bearer $USER_JWT_TOKEN" \
     -H "X-Stash-Authorization: Basic $jira_creds" \
     -d "
{
    \"card_requests\": [
        {
            \"tokens\": {
                \"pr_email_subject\": [
                    \"UFO/app-platform-server - Pull request #252: APF-767 Loosen up handling of connector error responses\"
                ]
            },
            \"connector_id\": \"$stash_connector_id\"
        }
    ]
}
" \
     "http://$HOST:$PORT/hero/cards/requests"
```

Where $stash_connector_id is the connector id you can get from discovery and $jira_creds is "user:password" base64 encoded.  Here is a script to prompt and output that (to prevent you from accidentally putting your password in bash history or something):

```
#!/bin/bash

read -p 'User: ' user
read -s -p 'Password: ' password
echo ''
echo -n "$user:$password" | base64
```

Remember that this blob of text is sensitive (your password), so if you store it in a file, don't share it and make sure the read permissions on it are strict.  If this ends up in an environment variable (or logs), be sure to clean up afterwards.

RyanBard: $ADMIN_JWT_TOKEN is tenant admin token
$USER_JWT_TOKEN is the user's token

I'll send links to our wiki page w/ dev tokens in chat.

RyanBard: You probably have to do something like this:

```java
    private Single<List<String>> getComments(final String baseUrl,
                                     final HttpHeaders headers,
                                     final StashPullRequest pullRequest) {
        final String url = String.format(StashConstants.STASH_ACTIVITIES_URL_FORMAT,
                baseUrl,
                pullRequest.getProjectKey(),
                pullRequest.getRepositorySlug(),
                pullRequest.getPullRequestId());

        final ListenableFuture<ResponseEntity<JsonDocument>> response = this.rest.exchange(
                url,
                HttpMethod.GET,
                new HttpEntity<>(headers),
                JsonDocument.class);

        return Async.toSingle(response)
                .map(ResponseEntity::getBody)
                .map(jsonDocument -> jsonDocument.<List<String>>read("$.values[*].comment.text"))
                .map(comments -> comments.stream().limit(COMMENTS_SIZE).collect(Collectors.toList()));
    }

    private Observable<Card> getCardsForStashPR(final HttpHeaders headers,
                                                final StashPullRequest pullRequest,
                                                final String baseUrl,
                                                final String routingPrefix) {
        logger.debug("Requesting pull request info from stash base url: {} and pull request info: {}", baseUrl, pullRequest);

        final Single<ResponseEntity<JsonDocument>> stashResponse = Async.toSingle(getPullRequestInfo(headers, pullRequest, baseUrl));
        final Single<List<String>> comments = getComments(baseUrl, headers, pullRequest);

        return Observable.zip(stashResponse.toObservable(), comments.toObservable(), Pair::of)
                .onErrorResumeNext(ObservableUtil::skip404)
                .map(pair -> convertResponseIntoCard(pair.getLeft(), pullRequest, routingPrefix, pair.getRight()));
    }
```

RyanBard: I have verified that Observable.zip is what you want to use.

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
RyanBard: Change this to something like:

```java
    private Single<String> getVersion(final HttpHeaders headers,
                              final String baseUrl,
                              final StashPullRequest pullRequest) {
        return getPullRequestInfo(headers, pullRequest, baseUrl)
                .map(ResponseEntity::getBody)
                .doOnSuccess(jsonDocument -> logger.debug("getVersion returned, jsonDoc={}", jsonDocument))
                .map(jsonDocument -> Integer.toString(jsonDocument.read("$.version")));
    }
```

```java
        return getVersion(headers, baseUrl, pullRequest)
                .flatMap(version -> {
                    final String url = String.format(StashConstants.STASH_PR_ACTION_FORMAT,
                            baseUrl,
                            pullRequest.getProjectKey(),
                            pullRequest.getRepositorySlug(),
                            pullRequest.getPullRequestId(),
                            stashAction.getAction(),
                            version);

                    // This header is added for skipping the CSRF check. Otherwise, Atlassian denies the pull request action.
                    headers.add(ATLASSIAN_TOKEN, "no-check");
...
```

connectors/stash/src/main/resources/static/discovery/metadata.hal
RyanBard: The client is not coping with multiple capture groups the way you want (they are sending whatever the last capture group is, so in this case, it would only send in the PR#).

You'll have to flag the inner ones as non-capturing (which will make your regex in the json different than the regex in the java code -- this also makes the regex unit test have to do what I mention in the other comment: expectList has to match what the client sends in).

Non-capturing group is specified with `?:`

```javascript
{
    "regex": "((?:[a-zA-Z0-9]+)/(?:[a-zA-Z0-9-]+) - Pull request #(?:[0-9]+):[ ])"
}
```

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
```diff
@@ -0,0 +1,444 @@
...
+                                  final StashPullRequest pullRequest) {
+        card.addAction(
+                new CardAction.Builder()
+                    .setLabel(this.cardTextAccessor.getActionLabel(STASH_COMMENTS))
+                    .setCompletedLabel(this.cardTextAccessor.getActionCompletedLabel(STASH_COMMENTS))
+                    .setActionKey(CardActionKey.USER_INPUT)
+                    .setUrl(buildActionUrl(routingPrefix, pullRequest, StashAction.COMMENTS))
+                    .addUserInputField(
+                            new CardActionInputField.Builder()
+                                .setId("Comment")
```
RyanBard: This doesn't match your comment action code.  The client will pass in:

```
--------------------
Req URL
--------------------
http://jbard-m01.local/connectors/c5f0d317-9770-42c2-bf53-4102145a5ea9//api/v1/UFO/app-platform-server/252/comments

--------------------
Req Headers
--------------------
Content-Type : application/x-www-form-urlencoded
Accept-Language : en;q=0.9, *;q=0.5
X-Request-ID : 3F289EA3-B25E-4BE5-A9A4-B3DCBA815DDD
Authorization : [S] Bearer e*I
x-stash-authorization : [S] Basic am*=


--------------------
Req Body
--------------------
Comment=Testing
```

Because the id is "Comment", however, down below, you're expecting "comment" (from the COMMENT_PARAM_KEY).  Use the constant for both places.

I don't think your other actions have the same problem because they don't have user input.

connectors/stash/src/test/resources/stash/responses/success.json
```diff
@@ -0,0 +1,171 @@
...
+      "header": {
+        "title": "Stash PR - Feature/APF-651 add image and test auth test/Sudharsan Sathiamoorthy",
+        "subtitle": "Description: * APF-651-Added integration test for getImage and getAuthKey\r\n* APF-651-Added integration test for test auth for connectors"
+      },
+      "body": {
+        "description": "* APF-651-Added integration test for getImage and getAuthKey\r\n* APF-651-Added integration test for test auth for connectors",
+        "fields": [
+          {
+            "type": "GENERAL",
+            "title": "Repository: ",
```
RyanBard: By the way, having the ": " in this string is causing the client to display things like: `Repository: :  app-platform-server`.

Don't specify the ": " in the text.properties file.

connectors/bitbucket-server/README.md
```diff
@@ -0,0 +1,7 @@
+# BitBucket Server Connector
```
rworsnop: It's "Bitbucket", not "BitBucket". Please replace all instances of this.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitBucketServerController.java
```diff
@@ -0,0 +1,435 @@
...
+
+        final ListenableFuture<ResponseEntity<JsonDocument>> response = this.rest.exchange(
+                url,
+                HttpMethod.GET,
+                new HttpEntity<>(headers),
+                JsonDocument.class);
+
+        final JsonDocument[] jsonDocuments = new JsonDocument[1];
+        Async.toSingle(response)
+                .subscribe(entity -> {
```
rworsnop: This is still broken. You never want to call `subscribe` in a connector. Spring does that for us further up the call stack . In the real, live connector, line 193 is going to be called before line 191.

You are getting comments asynchronously so it stands to reason that this method _cannot_ return `List<String>`. Instead it is going to have to return an Rx object, which will be incorporated into another RxJava pipeline.

Do let me know if you get stuck on this and I will suggest a solution.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitBucketServerController.java
```diff
@@ -0,0 +1,435 @@
...
+    }
+
+    private String getVersion(final HttpHeaders headers,
+                              final String baseUrl,
+                              final BitBucketServerPullRequest pullRequest) {
+
+        final ListenableFuture<ResponseEntity<JsonDocument>> response = getPullRequestInfo(headers, pullRequest, baseUrl);
+        final JsonDocument[] jsonDocuments = new JsonDocument[1];
+        Async.toSingle(response)
+                .subscribe(entity -> {
```
rworsnop: My other comment about calling `subscribe` applies here too.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitBucketServerController.java
SudharsanThumattiSathyamoorthy: I have done the changes to remove subscribe. I have not addressed all the pull requests comments. I have pushed only some of the changes. Apologies for the inconvenience caused.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitBucketServerController.java
SudharsanThumattiSathyamoorthy: I have done the changes to remove subscribe. I have not addressed all the pull requests comments. I have pushed only some of the changes. Apologies for the inconvenience caused.

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
```diff
@@ -0,0 +1,444 @@
...
+            produces = MediaType.APPLICATION_JSON_VALUE,
+            consumes = MediaType.APPLICATION_JSON_VALUE
+    )
+    public Single<ResponseEntity<Cards>> getCards(
+            @RequestHeader(AUTH_HEADER) final String authHeader,
+            @RequestHeader(BASE_URL_HEADER) final String baseUrl,
+            @RequestHeader(ROUTING_PREFIX) final String routingPrefix,
+            @Valid @RequestBody final CardRequest cardRequest) {
+
+        logger.trace("Cards requests for stash connector - authHeader: {}, baseUrlHeader: {}, routingPrefix: {}",
```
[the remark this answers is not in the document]
SudharsanThumattiSathyamoorthy: Removed authHeader from log message..

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
SudharsanThumattiSathyamoorthy: I have used ```UriComponentsBuilder``` instead of ```String.format``` in all the places..

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
SudharsanThumattiSathyamoorthy: Changed the code to above format.. Thanks for the suggestion... Since I have not tested the code initially, I did not realize the mistake... Tested it now and it works now...

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
SudharsanThumattiSathyamoorthy: The version changes for all the actions in Bitbucket.. To be on safer side, version is fetched everytime before executing an action..

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
SudharsanThumattiSathyamoorthy: I have used ```UriComponentsBuilder``` instead of ```String.format``` in all the places..

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
SudharsanThumattiSathyamoorthy: I have changed the code to above format..

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
SudharsanThumattiSathyamoorthy: I have used ```UriComponentsBuilder``` instead of ```String.format``` in all the places..

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
SudharsanThumattiSathyamoorthy: I have used ```UriComponentsBuilder``` instead of ```String.format``` in all the places..

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
SudharsanThumattiSathyamoorthy: Yeah.. Extracting the pull request email subject would be an easier option for the client..

connectors/stash/src/main/java/com/vmware/connectors/stash/utils/StashAction.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright © 2018 VMware, Inc. All Rights Reserved.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+package com.vmware.connectors.stash.utils;
+
+public enum StashAction {
+
+    // Retrieve all the comments from stash pull request.
```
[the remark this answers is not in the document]
SudharsanThumattiSathyamoorthy: I have changed it to javadoc comments..

connectors/stash/src/main/resources/static/discovery/metadata.hal
SudharsanThumattiSathyamoorthy: Changed the regex to ```(([a-zA-Z0-9]+)\/([a-zA-Z0-9-]+) - Pull request #([0-9]+):[ ])```..

connectors/stash/src/test/java/com/vmware/connectors/stash/StashControllerTest.java
```diff
@@ -0,0 +1,287 @@
...
+    public void testRegex() throws Exception {
+        final List<String> expectedList = ImmutableList.of(
+                // Project Name/ Repository Plug/ Pull request id.
+                "UFO/app-platform-server/244",
+                "UFO/app-platform-server/245",
+                "UFO/app-platform-server/241",
+                "UFO/app-platform-server/239",
+                "UFO/card-connectors/9");
+
+        mvc.perform(
```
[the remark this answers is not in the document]
SudharsanThumattiSathyamoorthy: I thought of using the ```testRegEx``` method in ```ControllerTestBase```.. Bitbucket regex involves three fields, but the logic inside the ```verifyRegex``` involves extracting only one field..

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
SudharsanThumattiSathyamoorthy: I have added ```test-auth``` api to BitbucketController to test basic auth..

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
SudharsanThumattiSathyamoorthy: I have rectified this error now.. Below is the generated comment url
```
https://hero/connectors/stash/api/v1/UFO/app-platform-server/246/comments
```

connectors/bitbucket-server/README.md
SudharsanThumattiSathyamoorthy: Changed ```BitBucket``` to ```Bitbucket``` in all the places..

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitBucketServerController.java
SudharsanThumattiSathyamoorthy: I have done the changes per @RyanBard suggestion.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitBucketServerController.java
SudharsanThumattiSathyamoorthy: I have done the changes per @RyanBard suggestion.

connectors/stash/src/main/java/com/vmware/connectors/stash/StashController.java
RyanBard: Yuck!  If they change the version on comments, approves, and rejects then yea we have to look it up before every action (the client isn't going to ask for a new card in-between the user commenting and then approving).

It sounds like Atlassian is a little too aggressive on their version checking :)

SudharsanThumattiSathyamoorthy:
-> Addressed all the pull request comments and ready for review.

-> Test all the Bitbucket actions manually and it is working.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/utils/BitbucketServerAction.java
```diff
@@ -0,0 +1,44 @@
...
+     * Decline an open bitbucket server pull request.
+     */
+    DECLINE("decline"),
+
+    /**
+     * Merge an approved bitbucket server pull request.
+     */
+    MERGE("merge");
+
+    private String action;
```
RyanBard: Make this final.

connectors/bitbucket-server/src/main/resources/static/discovery/metadata.hal
```diff
@@ -0,0 +1,9 @@
+{
+  "authorization_header": "x-bitbucket-server-authorization",
+  "backend_base_url_header": "x-bitbucket-server-base-url",
+  "fields": {
+    "pr_email_subject": {
+      "regex": "(([a-zA-Z0-9]+)\/([a-zA-Z0-9-]+) - Pull request #([0-9]+):[ ])"
```
RyanBard: Was this not a problem for you when you manually tested?  When I manually tested in the ios simulator, the client was sending in the `([0-9]+)` as the tokens.

RyanBard: I'll try again in with the updated code in a few minutes.

connectors/bitbucket-server/src/test/java/com/vmware/connectors/bitbucket/server/BitbucketServerControllerTest.java
```diff
@@ -0,0 +1,310 @@
...
+    @Test
+    public void testDiscovery() throws Exception {
+        testConnectorDiscovery();
+    }
+
+    @Test
+    public void testRegex() throws Exception {
+        final List<String> expectedList = ImmutableList.of(
+                // Project Name/ Repository Plug/ Pull request id.
+                "UFO/app-platform-server/244",
```
RyanBard: Again, the only reason you are having to reimplement all of this code is because your expectedList is populated with just the project/repository/id.  If you instead, populate it with what the client will actually send in:

```java
expectedList = ImmutableList.of(
        "UFO/app-platform-server - Pull request #244: ",
        "UFO/app-platform-server - Pull request #245: ",
        "UFO/app-platform-server - Pull request #241: ",
        "UFO/app-platform-server - Pull request #239: ",
        "UFO/card-connectors - Pull request #9: "
);
```

then you can simply call the common test regex method.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
```diff
@@ -0,0 +1,484 @@
...
+        final ListenableFuture<ResponseEntity<String>> response = this.rest.exchange(
+                uri,
+                HttpMethod.POST,
+                new HttpEntity<>(bitBucketServerComment, headers),
+                String.class);
+        return Async.toSingle(response);
+    }
+
+    @GetMapping("/test-auth")
+    @SuppressWarnings("PMD.JUnit4TestShouldUseTestAnnotation") // PMD assumes the method as Junit Test if the API url has 'test'.
```
RyanBard: You can get rid of this if you rename the method (don't start w/ test, so authTest or verifyAuth or something).

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
```diff
@@ -0,0 +1,484 @@
...
+    public Single<ResponseEntity<Void>> testAuth(@RequestHeader(AUTH_HEADER) final String authHeader,
+                                                 @RequestHeader(BASE_URL_HEADER) final String baseUrl) {
+        final HttpHeaders headers = new HttpHeaders();
+        headers.add(AUTHORIZATION, authHeader);
+
+        final ListenableFuture<ResponseEntity<JsonDocument>> response = this.rest.exchange(
+                baseUrl,
```
RyanBard: I suspect this either isn't going to work, or it can be simpler.  The client doesn't actually use this endpoint yet, so you'll have to test it with postman or curl.

What should happen is:

```
curl -H "Authorization: Bearer <vidmToken>" -H "X-Bitbucket-Servicer-Authorization: Basic <base64EncodeUserColonPassword>" 'http://localhost:port/test-auth'
```

Should return 200 if the base64EncodedUserColonPassword is correct, or 401 if not.

So if my user/pass was: user1/password1, then my x-auth header would be: "Basic dXNlcjE6cGFzc3dvcmQx" (encoding obtained by: `echo -n "user1:password1 | base64`) and I should expect a 200.

If I try that with a different password "Basic dXNlcjE6aW52YWxpZA==" (encoding obtained by: `echo -n "user1:invalid" | base64`), then I should expect a 401.

Since you are only requesting the baseUrl (instead of looking up a PR or something), you don't need the 404To200 conversion (in the Jira connector, that was to handle the situation where the test ticket id XYZ-123 didn't exist).

RyanBard: I confirmed that curling the base url does work (401 if creds are bad, 302 moved if creds are good, 302 moved if creds are blank).

I don't think the client will ever send in blank credentials "X-...-Authorization: Basic", so I'm ok w/ that returning 302.

We might want to map the 302 to 200, but it's probably fine as a 302 (we can revisit this when the client actually implements this).

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
```diff
@@ -0,0 +1,484 @@
...
+    }
+
+    private static ResponseEntity<Void> stripBody(final ResponseEntity<JsonDocument> entity) {
+        return ResponseEntity.status(entity.getStatusCode()).build();
+    }
+
+    private static Single<ResponseEntity<Void>> map404To200(final Throwable throwable) {
+        if (throwable instanceof HttpClientErrorException &&
+                HttpClientErrorException.class.cast(throwable).getStatusCode() == HttpStatus.NOT_FOUND) {
+            // If Bitbucket server is down, then we return 200 back to the client.
```
RyanBard: If the comment is accurate (I'm not sure it is), I wouldn't do this.  Let the client decide what to do (block the user from saving credentials or save it but realize they might want to check again later).

RyanBard:
Oh yea, I forgot to mention.  If manually testing actions on the example ios app, test with commit: efe4359f69f61c3a426fd9d62a4694a525b592a7

There is currently a bug preventing the cards' buttons from sending in the request.  It's being worked on, but isn't fixed yet, so efe4359f is the last good commit wrt the actions.

connectors/bitbucket-server/src/main/resources/static/discovery/metadata.hal
RyanBard:
```
Req Body
--------------------
{
  "card_requests" : [
    {
      "name" : "BitbucketServer",
      "tokens" : {
        "pr_email_subject" : [
          "252",
          "251"
        ]
      },
      "connector_id" : "7f66ed5c-e990-419c-b384-8d9084ba0dbc"
    }
  ]
}
```

While I think the client should return the first regex group (but it isn't specified in our docs what to do when multiple regex groups), it currently isn't (it returns the last regex group).  We can fix this by excluding the inner groups:

```
((?:[a-zA-Z0-9]+)\/(?:[a-zA-Z0-9-]+) - Pull request #(?:[0-9]+):[ ])
```

This will break your unit test as you currently have it written (that test is currently requiring the regex in the metadata.hal that we send to the client to be the same regex to support the same thing you are doing with our regex in the code that pulls out project/repo/id -- specified in a java constant).

RyanBard: The input in the ios example app that produced that (not correct) request:

```
* UFO/app-platform-server - Pull request #251: APF-730: Make misconfigured connectors unhealthy
* UFO/app-platform-server - Pull request #252: APF-767 Loosen up handling of connector error responses
```

RyanBard: When you ignore the inner groups, it will produce:

```
Req Body
--------------------
{
  "card_requests" : [
    {
      "name" : "BitbucketServer",
      "tokens" : {
        "pr_email_subject" : [
          "UFO/app-platform-server - Pull request #252: ",
          "UFO/app-platform-server - Pull request #251: "
        ]
      },
      "connector_id" : "7f66ed5c-e990-419c-b384-8d9084ba0dbc"
    }
  ]
}
```

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
SudharsanThumattiSathyamoorthy: Thanks for pointing it out.. Changed the comment to

```
If Bitbucket server base url is incorrect, then we return 200 back to the client.
```

If the base url is not correct, then it will throw 404. In this case, we are mapping it 200.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
SudharsanThumattiSathyamoorthy: If the base url is not correct, then it will throw 404. In this case, we are mapping it 200.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
SudharsanThumattiSathyamoorthy: Renamed the method name to ```verifyAuth```

connectors/bitbucket-server/src/test/java/com/vmware/connectors/bitbucket/server/BitbucketServerControllerTest.java
SudharsanThumattiSathyamoorthy: Changed the method as above. Now I am calling the common regex method to verify the result.

connectors/bitbucket-server/src/main/resources/static/discovery/metadata.hal
SudharsanThumattiSathyamoorthy: Yup!!

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/utils/BitbucketServerAction.java
SudharsanThumattiSathyamoorthy: Done.

connectors/jira/src/main/java/com/vmware/connectors/jira/JiraController.java
```diff
@@ -127,9 +127,8 @@ public JiraController(AsyncRestOperations rest, CardTextAccessor cardTextAccesso
     }
 
     @GetMapping("/test-auth")
-    @SuppressWarnings("PMD.JUnit4TestShouldUseTestAnnotation") // PMD assumes the method as Junit Test if the API url has 'test'.
```
SudharsanThumattiSathyamoorthy: Renamed the method here too.. Now we can get rid of this PMD violation..

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
RyanBard: Why should a misconfiguration report a 200 (the user's credentials are correct) when the client calls into test-auth?

connectors/bitbucket-server/src/main/resources/static/discovery/metadata.hal
RyanBard: Please change only the metadata.hal's regex (leave your java constant regex as-is) from:

```
"regex": "(([a-zA-Z0-9]+)\/([a-zA-Z0-9-]+) - Pull request #([0-9]+):[ ])"
```

to

```
"regex": "((?:[a-zA-Z0-9]+)\/(?:[a-zA-Z0-9-]+) - Pull request #(?:[0-9]+):[ ])"
```

so it will fix the ios simulator.

RyanBard: The regex in the metadata.hal is run by the client to produce that list of "pr_email_subject".  When you use postman, it is skipping what the client is doing with the metadata.hal.  This is fine most of the time, but like I say in the other comments, what is currently in the metadata.hal will cause the client to send in the wrong input (it only sends in the numbers).

You have to leave your java constant regex as it is for your java code to work, but the metadata.hal has to be changed for the client to work.  See the other comment for details.

RyanBard:
Just 2 things left and I'm satisfied (and I think @rworsnop's feedback is addressed, but he might want to take another look now):
* metadata.hal regex change to fix the client's behavior
* baseUrl 404 to 200 doesn't sound correct for test-auth

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
RyanBard: Don't worry about this one, I created APF-776 to followup on this (for both jira and bitbucket).

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
```diff
@@ -0,0 +1,483 @@
...
+        return headers;
+    }
+
+    private Single<String> getVersion(final HttpHeaders headers,
+                              final String baseUrl,
+                              final BitbucketServerPullRequest pullRequest) {
+
+        final ListenableFuture<ResponseEntity<JsonDocument>> response = getPullRequestInfo(headers, pullRequest, baseUrl);
+        return Async.toSingle(response)
+                .map(entity -> entity.getBody())
```
rworsnop: Can be a  method reference.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
```diff
@@ -0,0 +1,483 @@
...
+
+        final Single<ResponseEntity<JsonDocument>> bitBucketServerResponse = Async.toSingle(getPullRequestInfo(headers, pullRequest, baseUrl));
+        final Single<List<String>> comments = getComments(baseUrl, headers, pullRequest);
+
+        return Observable.zip(bitBucketServerResponse.toObservable(), comments.toObservable(), Pair::of)
+                .onErrorResumeNext(ObservableUtil::skip404)
+                .map(pair -> convertResponseIntoCard(pair.getLeft(), pullRequest, routingPrefix, pair.getRight()));
+    }
+
+    private ListenableFuture<ResponseEntity<JsonDocument>> getPullRequestInfo(final HttpHeaders headers,
```
rworsnop: The result of this is always converted to a `Single` by the caller. May as well just return a `Single`.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
```diff
@@ -0,0 +1,483 @@
...
+                                                                              final BitbucketServerPullRequest pullRequest,
+                                                                              final String baseUrl) {
+        final URI uri = UriComponentsBuilder.fromHttpUrl(baseUrl + "/rest/api/1.0/projects/{PROJECT_KEY}/repos/{REPOSITORY_SLUG}/pull-requests/{PULL_REQUEST_ID}")
```
rworsnop: Why not just use the `objects` parameter of the `exchange` method instead of building the URI like this? Ditto for the other examples.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
```diff
@@ -0,0 +1,483 @@
...
+                              final BitbucketServerPullRequest pullRequest) {
+
+        final ListenableFuture<ResponseEntity<JsonDocument>> response = getPullRequestInfo(headers, pullRequest, baseUrl);
+        return Async.toSingle(response)
+                .map(entity -> entity.getBody())
+                .map(jsonDocument -> Integer.toString(jsonDocument.read("$.version")));
+
+    }
+
+    private Observable<Card> getCardsForBitbucketServerPR(final HttpHeaders headers,
```
rworsnop: I have some ideas for improving the RxJava stuff here - please watch out for a PR requesting to merge into your branch. Basically I'm not a huge fan of returning an `Observable` when we know it only ever emits one item. On a related note, the method name refers to plural cards whereas there's only ever one. (I'll take care of that in my PR as well.)

RyanBard: We should create a followup ticket to fix the other connectors after the SingleUtil gets in.  I remember having to switch from single to observable and back a few times for some of this stuff.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
RyanBard: This can be cleaned up in a few other connectors as well (not in this PR).  We should create a ticket to fix the others.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
rworsnop: Please go ahead.

RyanBard: I created APF-782 for this.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
RyanBard: I created APF-783 for this.

RyanBard: The one thing that sucks about the `Object...` arg on `rest.exchange` is that they are positional params instead of named ones like these.  With only a handful or params, it's probably not too bad to use the positional params.

RyanBard: It looks like some of the overloads also have a Map for the uriVariables.  I remember having trouble getting that one to match what I wanted.  Maybe the overloads just got confusing and I gave up.

I'm not a fan of overloads btw :/

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
SudharsanThumattiSathyamoorthy: CHanged it to method reference.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
SudharsanThumattiSathyamoorthy: I have taken merged Rob's PR into my branch..

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
SudharsanThumattiSathyamoorthy: I am using the url directly in rest template and passing the date as objects paramater.

connectors/bitbucket-server/src/main/java/com/vmware/connectors/bitbucket/server/BitbucketServerController.java
SudharsanThumattiSathyamoorthy: I have pull your changes into my branch.

PR #24, opened 2018-02-01
Apf 699 Integrate with GB API for InstallApp action.

shsmysore:
- Multiple keywords for each app to trigger card request.
- Trigger native mdm app install using GreenBox api.
- Changed configuration file structure to take more data about managed apps.

As of now there is no support from GB Api to identify an app based on its bundleId. Hence platform specific app name is used to uniquely find GB app, to trigger app install.


connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/config/AppConfiguration.java
```diff
@@ -0,0 +1,64 @@
+/*
```
SudharsanThumattiSathyamoorthy: Please change the year to ```2018```.

connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/greenbox/GbApp.java
```diff
@@ -0,0 +1,37 @@
...
+ * Copyright © 2017 VMware, Inc. All Rights Reserved.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+package com.vmware.connectors.airwatch.greenbox;
+
+/**
+ * Created by harshas on 01/31/18.
+ */
+public class GbApp {
```
SudharsanThumattiSathyamoorthy: Can you please rename this class to ```GreenBoxApp``` ?  ```GbApp``` name might be too abstract.

connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/AirWatchController.java
```diff
@@ -174,12 +208,116 @@ public AirWatchController(AsyncRestOperations rest, CardTextAccessor cardTextAcc
...
+                .map(entity -> entity.getBody().read("$.eucToken"));
+    }
+
+    private Single<GbConnection> getGbConnection(URI gbBaseUri, String eucToken) {
+        logger.trace("getGbConnection called: GreenBox base url={}", gbBaseUri.toString());
+        return getCsrfToken(gbBaseUri, eucToken)
+                .map(csrfToken -> new GbConnection(gbBaseUri, eucToken, csrfToken));
+    }
+
+    private Single<ResponseEntity<HttpStatus>> installApp(String gbAppName, GbConnection gbSession) {
```
RyanBard: If you don't mind, try to name the 2 installApp methods differently.  It's easy to misread the code because of the overloading (especially when it's the same number of parameters and you aren't in an IDE).

connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/config/ManagedApp.java
```diff
@@ -0,0 +1,51 @@
...
+        return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, id);
```
RyanBard: Your hashCode and equals don't agree.

equals is only comparing id, but hashCode is comparing name and id.

The rule is for any 2 objects that are equal, they must have the same hashCode value.

When you break that rule, really bad things can happen with collections.  I can explain more and give you some examples in chat.

For now, either check name in your equals method also, or remove name from the hash.

It'd probably be better to just use Pojomatic (I'm not sure how you specify which properties to use, but I'm sure there is a way).

rworsnop: Yeah, use Pojomatic.

connectors/airwatch/src/main/service/rhel/files/default-conf/managed-apps.yml
```diff
@@ -1,16 +1,37 @@
-# Details about managed apps. Platform, app name and app id.
+# Details about managed apps.
 # You must override these details in /etc. By default there are no managed apps.
-android:
-
-ios:
+airwatch:
+  apps:
+    -
```
SudharsanThumattiSathyamoorthy: Currently, we are configuring apps manually in managed-apps.yml. This method will become cumbersome if the no of apps are more. We should provide an option in the console to configure apps and keywords and expose an API to save it in ```managed-apps.yml```. We can take this feature as an enhancement to ```airwatch-connector```.

connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/ConnConfiguration.java
```diff
@@ -0,0 +1,77 @@
...
+        return new AppConfigService(appConfigurations);
+    }
+
+    /*
+     * Make a set of app keywords and build regex for the connector.
+     * Replace regex place holder in the connector metadata file.
+     */
+    @Bean
+    public String connectorMetadata() throws IOException {
+
```
rworsnop: Please try this:
```
    public String connectorMetadata() throws IOException {

        final String connectorRegex = "(?i)" + appConfigurations.getApps().stream()
                .flatMap(appConfiguration -> appConfiguration.getKeywords().stream())
                .collect(Collectors.joining("|"));
        
        String metaData = IOUtils.toString(metadataHalResource.getInputStream(), Charset.defaultCharset());
        return metaData.replace("CONNECTOR_REGEX", connectorRegex);
    }
```

shsmysore: Thanks, It looks clean now.

connectors/airwatch/src/main/service/rhel/files/default-conf/managed-apps.yml
shsmysore: When no of managed apps increase, it will become cumbersome to supply details of all of them. 
I'm also wondering how many managed app might be there on an average. And also how frequently customers change managed app list in Airwatch console.

shsmysore:
Updated the doc with video demo.
https://confluence-euc.eng.vmware.com/pages/viewpage.action?pageId=104833444

PR #26, opened 2018-02-02
APF-786: Fix Bitbucket comment action

RyanBard:
I didn't notice that https://github.com/vmware/connectors-workspace-one/pull/21#discussion_r164250454 wasn't addressed, so the comments button wasn't working.

This fixes that issue.


SudharsanThumattiSathyamoorthy:
@RyanBard : Sorry about that... from that comment i was inferring about there was an extra backslash in the comment action url.. i fixed that.. i did not realise this part.. Thanks for fixing it..

PR #31, opened 2018-02-06
APF-777: Reduce Jira false positives

RyanBard:
* Adds unit tests to prove the false positives we were seeing will go away
* Adds a non-word char (outside the group) to prevent many false positives

### DIFF
--- a/common/connectors-test/src/main/java/com/vmware/connectors/test/ControllerTestsBase.java
+++ b/common/connectors-test/src/main/java/com/vmware/connectors/test/ControllerTestsBase.java
@@ -145,7 +145,7 @@ public class ControllerTestsBase {
 
         List<String> results = new ArrayList<>();
         for (String line : emailInput.split("\\n")) {
-            Matcher matcher = pattern.matcher(line);
+            Matcher matcher = pattern.matcher("\n" + line);
             while (matcher.find()) {
                 results.add(matcher.group(1));
             }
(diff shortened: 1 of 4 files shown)


common/connectors-test/src/main/java/com/vmware/connectors/test/ControllerTestsBase.java
```diff
@@ -145,7 +145,7 @@ private void verifyRegex(String regex, String emailInput, List<String> expected)
 
         List<String> results = new ArrayList<>();
         for (String line : emailInput.split("\\n")) {
-            Matcher matcher = pattern.matcher(line);
+            Matcher matcher = pattern.matcher("\n" + line);
```
RyanBard: The client code is doing something similar to this.

connectors/jira/src/test/java/com/vmware/connectors/jira/JiraControllerTests.java
```diff
@@ -73,6 +76,42 @@ public void testDiscovery() throws Exception {
...
+                "ABC-12", // not ideal, but not worth fixing
+                "ABC-13",
+                "ABC-14",
+                "ABC-15", // not ideal, but not worth fixing
+                "ABC-16",
+                "ABC-17",
+                "ABC-18",
+                "ABC-19", // not ideal, but not worth fixing
+//                "ABC-20", // Should not match due to leading "x"
+//                "D-2", // should not match
```
RyanBard: This D-2 through SA-2 is what the false positives were that were reported.  SA-2 was moved to APPSEC-2 (which made the false positive more confusing -- we didn't see APPSEC-2 in the email).

I started off with the false positives in the results (to make sure it was behaving as expected), then fixed the code and test, but left the commented out false positives to document what we're acknowledging the tradeoffs are.

PR #33, opened 2018-02-06
APF-784: Cleanup discovery details

RyanBard:
* Renames metadata.hal to metadata.json
* Updates discovery urls
* Updates tests


common/connectors-config/src/main/java/com/vmware/connectors/common/web/ConnectorRootController.java
```diff
@@ -53,7 +53,7 @@ public ConnectorRootController(@Value("${connector.hasTestAuth:false}") boolean
     }
```
SudharsanThumattiSathyamoorthy: In ```ConnectorsAutoConfiguration``` line no 96, we are setting the ```hal``` mime mapping to ```application/hal+json```. We have to change to ```application/json``` here.

RyanBard: It is correctly specifying content-type json for the json file (I assume spring has a default for json somewhere already) and correctly specifying hal+json for the actual hal content (I'm not sure if this is doing it or if something else in spring hateos is doing it).

It's possible I can remove this line, but I'm not sure yet.  I will try it and see.

SudharsanThumattiSathyamoorthy: Sorry, my bad. I was assuming this for only ```metadata.hal```. Its correct, no need to change this line.

PR #34, opened 2018-02-06
APF-762: Make root discovery cache configurable

RyanBard:

### DIFF
--- a/common/connectors-config/src/main/java/com/vmware/connectors/common/web/ConnectorRootController.java
+++ b/common/connectors-config/src/main/java/com/vmware/connectors/common/web/ConnectorRootController.java
@@ -22,7 +22,7 @@ import org.springframework.web.util.UriComponentsBuilder;
 
 import javax.servlet.http.HttpServletRequest;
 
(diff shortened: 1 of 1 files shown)


common/connectors-config/src/main/java/com/vmware/connectors/common/web/ConnectorRootController.java
```diff
@@ -31,10 +31,16 @@
 public class ConnectorRootController {
 
     private final boolean hasTestAuth;
+    private final long maxAge;
+    private final TimeUnit unit;
 
     @Autowired
-    public ConnectorRootController(@Value("${connector.hasTestAuth:false}") boolean hasTestAuth) {
+    public ConnectorRootController(@Value("${connector.hasTestAuth:false}") boolean hasTestAuth,
+                                   @Value("${rootDiscovery.cacheControl.maxAge:1}") long maxAge,
```
SudharsanThumattiSathyamoorthy: Its high time that we can start documenting the configuration parameters we are using in our repo.

PR #36, opened 2018-02-12
APF 770 change concur connector api from 3.0 to 2.1

SudharsanThumattiSathyamoorthy:
-> Downgraded Concur API version from 3.0 to 2.0.
-> Open, approve, reject actions are working fine.


connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
```diff
@@ -145,17 +164,31 @@ private String getConcurRequestTemplate(final String reason,
...
-        final ListenableFuture<ResponseEntity<JsonDocument>> result = this.rest.exchange("{baseUrl}/expense/reports/{id}",
+        final Single<ResponseEntity<JsonDocument>> result = getReportDetails(headers, id, baseUrl);
+
+        return result
+                .toObservable()
+                .onErrorResumeNext(ObservableUtil::skip404)
+                .map(entity -> convertResponseIntoCard(entity, baseUrl, id, routingPrefix));
+    }
+
+    private Single<ResponseEntity<JsonDocument>> getReportDetails(HttpHeaders headers, String id, String baseUrl) {
```
SudharsanThumattiSathyamoorthy: I was trying to find an API which supports ```HEAD```, could not find one.

connectors/bitbucket-server/src/main/resources/application.yml
```diff
@@ -2,4 +2,7 @@ server:
   compression:
      enabled: true
      mime-types: application/json,application/xml,text/html,text/xml,text/plain
-  server-header: bitbucket-server-connector
\ No newline at end of file
+  server-header: bitbucket-server-connector
+
```
SudharsanThumattiSathyamoorthy: I missed adding ```hasTestAuth``` to true for bitbucket connector.

connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
```diff
@@ -66,12 +67,13 @@
 
         final Set<String> expenseReportIds = cardRequest.getTokens(EXPENSE_REPORT_ID);
         if (CollectionUtils.isEmpty(expenseReportIds)) {
-            logger.debug("Expense report ids are empty for the base URL: {} ", baseUrl);
+            logger.info("Expense report ids are empty for the base URL: {} ", baseUrl);
```
RyanBard: Did you mean for this to stay info?

connectors/concur/src/main/resources/application.yml
```diff
@@ -3,3 +3,6 @@ server:
      enabled: true
      mime-types: application/json,application/xml,text/html,text/xml,text/plain
   server-header: concur-connector
+
+connector:
+  hasTestAuth: true
```
RyanBard: If Concur is using OAuth (not Http Basic Auth), then you shouldn't need a test-auth endpoint.  The test-auth endpoint is only useful because the client does not know whether the user typed in a bad user/password when prompted for Basic credentials.

With OAuth, the user is potentially prompted for user/password on the third party's website and given a screen to click Allow/Deny.  If the user messed up their credentials with OAuth, they would know about it immediately.

With Basic, we don't know if the credentials are good until the client tries to use them.  This is why the test-auth endpoint exists, so the client can check immediately instead of finding out when /requests/cards is called into.

connectors/concur/src/test/java/com/vmware/connectors/concur/ConcurControllerTests.java
```diff
@@ -246,17 +289,18 @@ private void testRequestCards(final String requestFile,
     }
 
     private ResponseActions expect(final String issue) {
-        return this.mockConcur.expect(requestTo("https://concursolutions.com/expense/reports/" + issue))
+        return this.mockConcur.expect(requestTo("https://implementation.concursolutions.com/api/expense/expensereport/v2.0/report/" + issue))
                 .andExpect(method(GET))
-                .andExpect(MockRestRequestMatchers.header(AUTHORIZATION, "Bearer concur-token"));
+                .andExpect(MockRestRequestMatchers.header(AUTHORIZATION, "OAuth 0_xxxxEKPk8cnYlWaos22OpPsLk="))
+                .andExpect(MockRestRequestMatchers.header(ACCEPT, APPLICATION_JSON_VALUE));
```
RyanBard: Does the 2.0 endpoint really support JSON?  I only see XML mentioned in the docs: https://developer.concur.com/api-reference/expense/expense-report/expense-report-get.html

connectors/concur/src/test/resources/concur/responses/report_id_1.json
```diff
@@ -1,193 +1,309 @@
...
+  "Custom17": {},
+  "Custom18": {},
+  "Custom19": {},
+  "Custom20": {},
+  "LedgerName": "ORACLE",
+  "PolicyID": "gWk3QXwsDWJoW5Yar$ph$sJI$pPLdwGyGTM71g",
+  "EverSentBack": "N",
+  "ReportImageURL": "https://imagingupload.concursolutions.com/file/t0030425xl9t/52C2430FA25414D35216B02E93D2C8F5011FA306E2CBCFFEEE2A53653F4A0F9C26BA34BAC9FC8CFCA12F323D75B82A90246F63CFDB3C1CDCF1573E06E6E0B92FH722CCBEF3F16D8B8C7743AC35E44941D?id=7BCBC522BBF94102B368&e=t0030425xl9t&t=AN",
+  "HasException": "N",
+  "WorkflowActionURL": "https://implementation.concursolutions.com/api/expense/expensereport/v1.1/report/gWujNPAb67r9LjhqgN7BEYYaQOWzavXBtUP1sej$sXfPQ/WorkFlowAction",
```
RyanBard: Is this URL from a real response?  The WorkFlowAction after the reportId in the URL isn't mentioned in their docs: https://developer.concur.com/api-reference-deprecated/version-one-one/expense-report/expense-report-header-post.html

SudharsanThumattiSathyamoorthy: The above URL is not the correct documentation. Documentation link for approving/rejecting the workflowaction is "https://developer.concur.com/api-reference/expense/expense-report/post-report-workflow-action.html"

connectors/concur/src/test/java/com/vmware/connectors/concur/ConcurControllerTests.java
SudharsanThumattiSathyamoorthy: Yes, it supports json. By default, it will xml reponse. We have to add "Accept - application/json" in the request header to receive the json reponse. I have created a sample report for testing. Below is the curl command we can use to verify the response:

```
curl -X GET \
https://implementation.concursolutions.com/api/expense/expensereport/v2.0/report/F66E2EC1743944E9929F \
  -H 'accept: application/json' \
  -H 'authorization: OAuth 0_5sfyEKPk8cnYlWaos22OpPsLk='
```

connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
```diff
@@ -83,51 +86,54 @@
...
-                HttpMethod.POST,
-                new HttpEntity<>(concurRequestTemplate, headers),
-                ConcurResponse.class,
-                baseUrl,
-                workflowstepId);
-
-        logger.info("Concur response : {} and the response code is : {} ", response.get().getBody(), response.get().getStatusCode());
-        return Async.toSingle(response)
-                .map(entity -> status(entity.getStatusCode()).build());
+        return getWorkFlowActionUrl(headers, reportID, baseUrl)
```
rworsnop: This is more concise:
```
        return getWorkFlowActionUrl(headers, reportID, baseUrl)
                .flatMap(workflowActionUrl -> Async.toSingle(rest.postForEntity(workflowActionUrl,
                         new HttpEntity<>(concurRequestTemplate, headers),
                        String.class,
                        baseUrl,
                        reportID)))
                .map(entity -> status(entity.getStatusCode()).build());
```

connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
```diff
@@ -145,74 +151,81 @@ private String getConcurRequestTemplate(final String reason,
...
+    }
 
-        return Async.toSingle(result)
-                .toObservable()
-                .onErrorResumeNext(ObservableUtil::skip404)
-                .map(entity -> convertResponseIntoCard(entity, baseUrl, id, routingPrefix));
+    private Single<String> getWorkFlowActionUrl(final HttpHeaders headers,
+                                                final String id,
+                                                final String baseUrl) {
+        final Single<ResponseEntity<JsonDocument>> result = getReportDetails(headers, id, baseUrl);
```
rworsnop: This is more concise:
```
      return getReportDetails(headers, id, baseUrl)
                .map(ResponseEntity::getBody)
                .map(jsonDocument -> jsonDocument.read("$.WorkflowActionURL"));
```

connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
SudharsanThumattiSathyamoorthy: Changes the code to above format.

connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
SudharsanThumattiSathyamoorthy: I have changed the code to above format.

PR #47, opened 2018-02-20
APF-870: Add NodeJS ServiceNow connector

RyanBard:


rworsnop:
Closing this PR but let's leave the branch alone. In the future we might want to reorganize the whole repo to have Java as a sibling of Node. Also there is common code here.

PR #49, opened 2018-02-20
APF-871: Change subtitle from String to List

RyanBard:
* Changes subtitle from String to List<String>
* Changes AWS Cert connector to move domain name into subtitle
* Reduces repeated title info in Github PR connector
* Reduces repeated title info in Gitlab MR connector
* Changes ServiceNow connector to move req number into subtitle


RyanBard:
example of what is fixed by this:

screen shot 2018-02-20 at 4 22 28 pm


now becomes:

jira-collapsed

PR #51, opened 2018-02-22
APF-860 ReadMe and http 500-exception cases.         Signed-off-by: S…

shsmysore:
- Prevent 500 error exceptions.
- Updated ReadMe.

Signed-off-by: Shree Harsha <harshas@vmware.com>

### DIFF
--- a/connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/AirWatchController.java
+++ b/connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/AirWatchController.java
@@ -31,7 +31,6 @@ import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
-import org.springframework.util.CollectionUtils;
 import org.springframework.util.concurrent.ListenableFuture;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.GetMapping;
@@ -121,13 +120,14 @@ public class AirWatchController {
 
         if (StringUtils.isAnyBlank(udid, clientPlatform)) {
             logger.debug("Either device UDID or client platform is blank.");
-            return Single.just(new ResponseEntity<>(BAD_REQUEST));
+            return Single.just(ResponseEntity.badRequest().build());
         }
 
         Set<String> appKeywords = cardRequest.getTokens("app_keywords");
 
-        if (CollectionUtils.isEmpty(appKeywords)) {
-            return Single.just(ResponseEntity.ok(new Cards()));
+        if (appKeywords == null) {
+            logger.debug("Request is missing app_keywords token.");
+            return Single.just(ResponseEntity.badRequest().build());
         }
 
         HttpHeaders headers = new HttpHeaders();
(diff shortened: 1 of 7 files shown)


connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/AirWatchController.java
```diff
@@ -115,8 +118,18 @@ public AirWatchController(AsyncRestOperations rest, CardTextAccessor cardTextAcc
 
         String udid = cardRequest.getTokenSingleValue(UDID_KEY);
         String clientPlatform = cardRequest.getTokenSingleValue(PLATFORM_KEY);
+
+        if (StringUtils.isAnyBlank(udid, clientPlatform)) {
+            logger.debug("Either device UDID or client platform is blank.");
+            return Single.just(new ResponseEntity<>(BAD_REQUEST));
```
rworsnop: Let's be consistent: `ResponseEntity.badRequest()`

connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/AirWatchController.java
```diff
@@ -115,8 +118,18 @@ public AirWatchController(AsyncRestOperations rest, CardTextAccessor cardTextAcc
...
+        }
+
         Set<String> appKeywords = cardRequest.getTokens("app_keywords");
 
+        if (CollectionUtils.isEmpty(appKeywords)) {
```
rworsnop: Is this really necessary? Looks like the code below ought to be able to handle empty `appKeywords` just fine.

connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/AirWatchController.java
```diff
@@ -115,8 +118,18 @@ public AirWatchController(AsyncRestOperations rest, CardTextAccessor cardTextAcc
...
+            return Single.just(ResponseEntity.ok(new Cards()));
+        }
+
         HttpHeaders headers = new HttpHeaders();
         headers.set(AUTHORIZATION, awAuth);
```
SudharsanThumattiSathyamoorthy: In line 161, we are fetching the authorization token by splitting the string based on ```(?i)Bearer```. Incase, if the client send an authorization token starts with ```Basic```, then it will result in ```ArrayIndexOutOfBoundsException```. 

We can avoid this by checking whether the ```authToken``` begins with ```Baisc`` or ```Bearer``` and we can retrieve the token accordingly. If it is not any of the above values, throw an ```InvalidAuthException``` back to the client.

connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/AirWatchController.java
shsmysore: Removed `CollectionUtils` and added a check to make sure this token exists in the request.

connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/AirWatchController.java
shsmysore: But since we support only Bearer type, this will result in 401.

PR #57, opened 2018-03-06
APF-907-Added allow_repeated actions for jira and concur connector.

SudharsanThumattiSathyamoorthy:


common/core/src/main/java/com/vmware/connectors/common/payloads/response/CardAction.java
```diff
@@ -167,12 +170,22 @@ public boolean isRemoveCardOnCompletion() {
...
      * @return the completed label
      */
     public String getCompletedLabel() {
         return completedLabel;
     }
 
+    /**
+     * Returns "allow repeated" value. If set to true, then the client can show
+     * the "OPEN_IN" action irrespective of "completed label".
+     *
```
shsmysore: We need to fix this comment. `OPEN_IN` is just an example.
But connectors developers can use the property for any of their card-action.

RyanBard: Yes, that's correct.

SudharsanThumattiSathyamoorthy: I have made the comment generic now.

PR #58, opened 2018-03-06
APF-906-Made completed lable mandatory in card connector schema. Fixe…

SudharsanThumattiSathyamoorthy:
…d all connectors where completed label is missing.

### DIFF
--- a/common/core/src/main/resources/schemata/herocard-connector-response-schema.json
+++ b/common/core/src/main/resources/schemata/herocard-connector-response-schema.json
@@ -77,7 +77,7 @@
         "request":    {"type": "object"},
         "allow_repeated" : {"type": "boolean"}
       },
-      "required": ["id", "completed_label"],
+      "required": ["id", "completed_label", "url", "action_key", "label", "type"],
       "additionalProperties": false
     },
 
--- a/common/core/src/main/resources/schemata/herocard-connector-response-schema.json
+++ b/common/core/src/main/resources/schemata/herocard-connector-response-schema.json
@@ -76,6 +76,7 @@
         "user_input": {"type": "array", "items": {"$ref": "#/definitions/CARD_ACTION_USER_INPUT"}},
         "request":    {"type": "object"}
       },
+      "required": ["id", "completed_label"],
       "additionalProperties": false
     },
 
--- a/connectors/airwatch/src/main/resources/cards/text.properties
+++ b/connectors/airwatch/src/main/resources/cards/text.properties
@@ -1,3 +1,5 @@
 header={0}
 body=Please install this application for better interaction with the resource.
+installApp=Install
 installApp.label=Install
+installApp.completedLabel=Installed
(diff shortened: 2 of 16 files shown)


connectors/airwatch/src/main/resources/cards/text.properties
```diff
@@ -1,3 +1,5 @@
 header={0}
 body=Please install this application for better interaction with the resource.
+installApp=Install
 installApp.label=Install
+installApp.completedLabel=Installed
```
shsmysore: I'm just wondering if we need to make it something like "Install triggered" or "Installing" ?
Because I have always seen it takes considerable amount of time to reach "Installed" state.

SudharsanThumattiSathyamoorthy: If we keep the text as ```Install triggered```, then everytime the user opens the email, it will show the state as ```Install triggered``` or ```Installing```, which might be misleading to the user.

common/core/src/main/resources/schemata/herocard-connector-response-schema.json
```diff
@@ -76,6 +76,7 @@
         "user_input": {"type": "array", "items": {"$ref": "#/definitions/CARD_ACTION_USER_INPUT"}},
         "request":    {"type": "object"}
       },
+      "required": ["id", "completed_label"],
```
RyanBard: Actually, could you make this `["id", "action_key", "label", "completed_label", "url", "type"]`

https://github.com/vmwaresamples/card-connectors-guide/pull/7/files

connectors/airwatch/src/main/resources/cards/text.properties
RyanBard: I don't like "Installing" (makes it confusing between whether the action is being communicated with the server or not).

"Install triggered", maybe, but it uses up a lot of space.  Maybe run this one by David and see what he thinks.

RyanBard: I would proceed w/ "Installed" for now, and I'll talk to David today (I actually have a meeting wrt the AirWatch connector today).

If he likes "Install triggered" better, we can change it later.

common/core/src/main/resources/schemata/herocard-connector-response-schema.json
SudharsanThumattiSathyamoorthy: Added ```"id", "action_key", "label", "completed_label", "url", "type"]``` fields to required.

RyanBard: I don't see them.  Did you lose the changes when pulling/rebasing or something?

SudharsanThumattiSathyamoorthy: sorry.. i forgot to push the changes...  I have pushed the changes now..

connectors/airwatch/src/main/resources/cards/text.properties
SudharsanThumattiSathyamoorthy: Sure... Please let me know about the completed label after the discussion...

RyanBard: I forgot to follow up on this.  We didn't have the meeting yet (had to focus on other things -- verifying the connector in multiple environments).

We never discussed the install button, however, there was the start of a discussion on the UX ticket for this connector (but it never really got anywhere).

At the end of last week, I mentioned the install/installation-triggered thing to Evan and he said we can discuss it at a later time (presumably after this release).  When we do, we'll be discussing the life cycle of our cards (when are they fetched?  when are they retried?  can they be refreshed?  etc.) and how to best communicate that to our users with our UI.

For now, I've added you as a watcher APF-816 (the AirWatch connector UX ticket) in case this discussion happens again before that.

PR #62, opened 2018-03-10
APF-919 Convert connectors to WebClient

rworsnop:
AsyncRestTemplate is deprecated in 5.0.
Added drop-in replacement for MockRestServiceServer for use
while we wait for Spring to support WebClient.
Re-organized to use arg list to propagate Locale instead of Context,
which has now been removed.


connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
```diff
@@ -39,54 +37,54 @@
...
+    private final WebClient rest;
+    private final CardTextAccessor cardTextAccessor;
+    private final Resource concurrRequestTemplate;
 
-    @Value("classpath:static/templates/concur-request-template.xml")
-    private org.springframework.core.io.Resource concurrRequestTemplate;
+    @Autowired
+    public ConcurController(WebClient rest,
+                            CardTextAccessor cardTextAccessor,
+                            @Value("classpath:static/templates/concur-request-template.xml") Resource concurrRequestTemplate) {
```
SudharsanThumattiSathyamoorthy: Typo: ```concurRequestTemplate```.

PR #63, opened 2018-03-12
APF-906-Added completed_label for aws cert action. 

SudharsanThumattiSathyamoorthy:

### DIFF
--- a/common/core/src/main/resources/schemata/herocard-connector-response-schema.json
+++ b/common/core/src/main/resources/schemata/herocard-connector-response-schema.json
@@ -76,7 +76,9 @@
         "type":       {"$ref": "#/definitions/HTTP_METHODS"},
         "user_input": {"type": "array", "items": {"$ref": "#/definitions/CARD_ACTION_USER_INPUT"}},
         "request":    {"type": "object"},
-        "allow_repeated" : {"type": "boolean"}
+        "allow_repeated" : {"type": "boolean"},
+        "remove_card_on_completion": {"type": "boolean"},
+        "primary": {"type": "boolean"}
       },
       "required": ["id", "completed_label", "url", "action_key", "label", "type"],
       "additionalProperties": false
--- a/connectors/aws-cert/src/main/java/com/vmware/connectors/aws/cert/AwsCertController.java
+++ b/connectors/aws-cert/src/main/java/com/vmware/connectors/aws/cert/AwsCertController.java
@@ -259,7 +259,7 @@ public class AwsCertController {
         CardAction.Builder dismissAction = CardAction.Builder
                 .dismissAction()
                 .setLabel(cardTextAccessor.getActionLabel("dismiss"))
-                .setCompletedLabel(cardTextAccessor.getActionCompletedLabel("dismiss"));
+                .setCompletedLabel(cardTextAccessor.getActionLabel("dismiss"));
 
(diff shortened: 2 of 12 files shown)


connectors/aws-cert/src/main/java/com/vmware/connectors/aws/cert/AwsCertController.java
```diff
@@ -258,7 +258,8 @@ private Card makeCard(
 
         CardAction.Builder dismissAction = CardAction.Builder
                 .dismissAction()
-                .setLabel(cardTextAccessor.getActionLabel("dismiss"));
+                .setLabel(cardTextAccessor.getActionLabel("dismiss"))
+                .setCompletedLabel(cardTextAccessor.getActionCompletedLabel("dismiss"));
```
RyanBard: In my opinion, just use the default (even if we can't specify locale).  They will never really see this completed label.  The whole point of the dismiss button is to make the card go away.

common/core/src/main/resources/schemata/herocard-connector-response-schema.json
```diff
@@ -76,7 +76,9 @@
         "type":       {"$ref": "#/definitions/HTTP_METHODS"},
         "user_input": {"type": "array", "items": {"$ref": "#/definitions/CARD_ACTION_USER_INPUT"}},
         "request":    {"type": "object"},
-        "allow_repeated" : {"type": "boolean"}
+        "allow_repeated" : {"type": "boolean"},
```
SudharsanThumattiSathyamoorthy: Added ```remove_card_on_completion``` and ```primary``` to the ```hero_card_schema```.

connectors/aws-cert/src/main/java/com/vmware/connectors/aws/cert/AwsCertController.java
SudharsanThumattiSathyamoorthy: Modified the ```completedLabel``` to have normal action label. Changed the test cases accordingly.

connectors/aws-cert/src/main/java/com/vmware/connectors/aws/cert/AwsCertController.java
```diff
@@ -259,7 +259,7 @@ private Card makeCard(
         CardAction.Builder dismissAction = CardAction.Builder
                 .dismissAction()
                 .setLabel(cardTextAccessor.getActionLabel("dismiss"))
-                .setCompletedLabel(cardTextAccessor.getActionCompletedLabel("dismiss"));
+                .setCompletedLabel(cardTextAccessor.getActionLabel("dismiss"));
```
RyanBard: You shouldn't even need this line (you can let it default to "Completed"), but it's not worth holding this PR up.

SudharsanThumattiSathyamoorthy: If ```completed_label``` is missing the card response, then the test cases will fail since it is a mandatory field in ```hero_card_schema.json```.

RyanBard: https://github.com/vmware/connectors-workspace-one/blob/5b545876c0e7484fe8a799eb63384900ed4e4eeb/common/core/src/main/java/com/vmware/connectors/common/payloads/response/CardAction.java#L73

PR #68, opened 2018-03-16
Apf 895 add support for image link in card json schema

SudharsanThumattiSathyamoorthy:


common/connectors-config/src/main/java/com/vmware/connectors/common/utils/CommonUtils.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Common utility functions to be used across connectors parent.
+ */
+public final class CommonUtils {
+
+    public static final String IMAGE_PATH = "/images/connector.png";
+
+    private CommonUtils() {
+        // Singleton class.
```
RyanBard: It's a utility class.  A singleton has exactly 1 instance.  This has no instances (all static methods, no need for an instance).

common/connectors-config/src/main/java/com/vmware/connectors/common/utils/CommonUtils.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public static void buildImageUrl(final Card.Builder card, final HttpServletRequest request) {
```
RyanBard: Please rework this to reduce duplication and allow a better contract (you can keep the Card.Builder one as an overload if you want):
* https://github.com/vmware/connectors-workspace-one/blob/cc60cf58be7828d8ab24c8a9c52bff0620ef0103/common/connectors-config/src/main/java/com/vmware/connectors/common/web/ConnectorRootController.java#L74

A better approach is:

```java
public static String buildImageUrl(HttpServletRequest request) {
    return buildImageUrl(request, DEFAULT_IMAGE_PATH);
}

public static String buildImageUrl(HttpServletRequest request, String imagePath) {
    return UriComponentBuilder.fromHttpRequest(request).path(imagePath).build().toUriString();
}

// if you want to keep the card builder overload
public static void buildImageUrl(Card.Builder card, HttpServletRequest request) {
    card.setImageUrl(buildImageUrl(request));
}


// and in the Discovery code:
if (imageResource.exists()) {
    String image = CommonUtils.buildImageUrl(request);
    resource.add(new Link(image, "image"));
}
```

That is more flexible (what if someone wants an image url for something that doesn't involve building a card -- like the discovery code) and doesn't force the caller to have to break up their builder method chains if they don't want to.

SudharsanThumattiSathyamoorthy: I have added overloaded methods for ```buildConnectorImageUrl```. We can't use ```UriComponentBuilder.fromHttpRequest``` here as the original request was ```/card/requests```. Image url will be ```http://bitbucket-connector/card/requests/images/connector.png```. I retained the logic to build the image url.

common/connectors-config/src/main/java/com/vmware/connectors/common/utils/CommonUtils.java
SudharsanThumattiSathyamoorthy: Modified the comment to ```Utility class.```

common/connectors-config/src/main/java/com/vmware/connectors/common/utils/CommonUtils.java
RyanBard: You can use fromHttpRequest and replacePath.  There is an example in the mock-connectors repo I believe:

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/util/UriComponentsBuilder.html#replacePath-java.lang.String-

connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/AirWatchController.java
```diff
@@ -216,6 +233,10 @@ public AirWatchController(WebClient rest, CardTextAccessor cardTextAccessor,
                 .setHeader(cardTextAccessor.getHeader(locale, appName))
                 .setBody(cardBodyBuilder.build())
                 .addAction(appInstallActionBuilder.build());
+
+        // Set Image url.
+        CommonUtils.buildConnectorImageUrl(cardBuilder, request);
```
RyanBard: We don't want to set this currently.

Notice that AirWatch connector does not have an image right now:
* https://github.com/vmware/connectors-workspace-one/tree/master/connectors/airwatch/src/main/resources/static

SudharsanThumattiSathyamoorthy: I have removed ```buildConnectorImageUrl``` from the Airwatch connector.

common/connectors-config/src/main/java/com/vmware/connectors/common/utils/CommonUtils.java
SudharsanThumattiSathyamoorthy: Replaced the logic using ```UriComponentsBuilder.fromHttpRequest(request).replacePath(path).build().toString()```.

connectors/salesforce/src/test/java/com/vmware/connectors/salesforce/SalesforceControllerTest.java
```diff
@@ -22,13 +22,15 @@
...
 import org.junit.jupiter.params.provider.ValueSource;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.core.io.Resource;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.test.web.client.ResponseActions;
 import org.springframework.test.web.client.match.MockRestRequestMatchers;
 import org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder;
+import org.springframework.web.client.AsyncRestTemplate;
```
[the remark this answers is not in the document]
SudharsanThumattiSathyamoorthy: I was comparing all the connector tests with ```master``` branch and was reverting some of my changes... I guess this import got in by mistake...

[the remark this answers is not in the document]
SudharsanThumattiSathyamoorthy: I have merged ```APF-895-add-support-for-image-link-in-card-json-schema-suggestion``` into my branch.

PR #72, opened 2018-03-19
APF-957 Tidy up reactive stuff

rworsnop:
Used Mono where possible and got rid of
Reactive.toResponseEntity


connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
```diff
@@ -133,15 +133,14 @@ private String getConcurRequestTemplate(final String reason,
     }
 
 
-    private Flux<Card> getCardsForExpenseReport(final String authHeader,
+    private Mono<Card> getCardForExpenseReport(final String authHeader,
```
RyanBard: This is (1 of the many reasons) why I don't IntelliJ's default indention of function params.

The indentation is off again (by the 1 char difference).

PR #73, opened 2018-03-20
APF-980: Improve connectors' regex

RyanBard:
* Adds word boundaries to AirWatch connector's regex
* Adds word boundaries to ServiceNow connector's regex
* Changes Jira connector's regex from non-word char to word boundary
* Changes regex test code to do what the Android client implemented
* Adds more tests for AirWatch connector's regex


RyanBard:
I updated the AirWatch, Jira, and ServiceNow connectors' regular expressions.

The Salesforce regex was complicated enough that I didn't want to mess with it (it already has a few false positives unrelated to word boundaries).

Most of the other connectors are specific enough that false positives don't seem to be a problem, so I didn't add any complexity to their regular expressions.

connectors/servicenow/src/test/java/com/vmware/connectors/servicenow/ServiceNowControllerTest.java
```diff
@@ -74,26 +74,26 @@ void testRegex() throws Exception {
         List<String> expected = Arrays.asList(
                 "REQ0010001",
                 "REQ0010003",
-                "REQ0010000", // not ideal
+//                "REQ0010000", // Should not match due to too many numbers
                 "REQ0010005",
                 "REQ0010007",
                 "REQ0010009",
-                "REQ0010010", // not ideal
+                "REQ0010010", // Should not match due to trailing "X"
```
RyanBard: Somehow, I didn't finish commenting these out and didn't notice that I hadn't re-run the build since an earlier version of the code.

I had @rworsnop go ahead and revert instead of me rushing out the fix, so I can verify manually again (in case I fooled myself on which version of the code I tested in my vm).

PR #74, opened 2018-03-20
Apf 980 improve connectors regex

RyanBard:
* Adds word boundaries to AirWatch connector's regex
* Adds word boundaries to ServiceNow connector's regex
* Changes Jira connector's regex from non-word char to word boundary
* Changes regex test code to do what the Android client implemented
* Adds more tests for AirWatch connector's regex


RyanBard:
Sorry about that.

I tested local build, a build in the VM (w/o skipping tests/pmd), and manually verified the connectors again double checking that I was running the code I thought I was.

Should be good now.

PR #77, opened 2018-04-03
Apf 1009 Fix app discovery duplicate cards

shsmysore:


connectors/airwatch/src/test/resources/connector/requests/requestDuplicate.json
```diff
@@ -0,0 +1,7 @@
+{
+  "tokens": {
+    "udid": ["ABCD"],
+    "platform": ["android"],
+    "app_keywords": ["Boxer", "email app", "Concur"]
```
SudharsanThumattiSathyamoorthy: Please correct me if I am wrong. We are adding three unique app keywords here. Can you please add any one of the app names again in the app keywords and check if the duplicate cards were not generated in the ```response.json``` file.

RyanBard: If I understand correctly, that is what it is doing.  So it's requesting boxer, boxer (email app), and concur and using the same response as the other tests (boxer + concur returned).

RyanBard: It looks like concur is excluded for other reasons, so it's verifying that only 1 card is returned (boxer).

shsmysore: Right, both the keywords resolve to the same `ManagedApp`.

The mock device service says Concur app is installed, Boxer app is not installed. (Only once for each app). Hence if connector tries to ask the install status twice, test would fail.

PR #84, opened 2018-04-10
APF-1045: heapdump on OutOfMemoryErrors

RyanBard:
* Moves systemd ExecStart command to a script for easier maintenance
* Adds options to dump heap on OutOfMemoryError
* Updates memory options to be more reasonable than JVM defaults

### DIFF
--- a/connectors/airwatch/src/main/service/rhel/files/default-conf/jvm.conf
+++ b/connectors/airwatch/src/main/service/rhel/files/default-conf/jvm.conf
@@ -1,0 +1,3 @@
+export JVM_OPTS=""
+export JVM_OPTS="$JVM_OPTS -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp"
+export JVM_OPTS="$JVM_OPTS -XX:+UnlockExperimentalVMOptions -XX:MaxRAMFraction=2"
--- a/connectors/airwatch/src/main/service/rhel/files/default-conf/start.sh
+++ b/connectors/airwatch/src/main/service/rhel/files/default-conf/start.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+connector_name='airwatch'
+install_path="/opt/vmware/connectors/$connector_name"
+etc_path="/etc/opt/vmware/connectors/$connector_name"
+
+source $install_path/jvm.conf
+
+/usr/bin/java $JVM_OPTS \
+    -jar $install_path/${connector_name}-connector.jar \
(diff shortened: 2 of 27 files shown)


connectors/airwatch/src/main/service/rhel/files/default-conf/jvm.conf
```diff
@@ -0,0 +1,3 @@
+export JVM_OPTS=""
```
RyanBard: The jvm.conf files in this PR are all identical.

connectors/airwatch/src/main/service/rhel/files/default-conf/start.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+connector_name='airwatch'
```
RyanBard: The airwatch start.sh is a little different than the others (it has the managed-apps.yml file as well as the 2 application.properties files).

connectors/aws-cert/src/main/service/rhel/files/default-conf/start.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+connector_name='aws-cert'
```
RyanBard: All of the start.sh files except airwatch-connector's are the same except this 1 connector_name variable.

connectors/airwatch/src/main/service/rhel/files/systemd/airwatch-connector.service
```diff
@@ -5,7 +5,7 @@ After=network.target
 [Service]
 Type=simple
 User=roswell
-ExecStart=/usr/bin/java -jar /opt/vmware/connectors/airwatch/airwatch-connector.jar  --spring.config.additional-location=file:/opt/vmware/connectors/airwatch/application.properties,file:/etc/opt/vmware/connectors/airwatch/application.properties,file:/etc/opt/vmware/connectors/airwatch/managed-apps.yml
+ExecStart=/bin/bash /opt/vmware/connectors/airwatch/start.sh
```
RyanBard: I don't think systemd allowed me to just run the script (I had to specify /bin/bash), however, I'll try this again now that I have things working (something else might have been happening yesterday that confused me).

RyanBard: And I see some notes that I forgot to follow up on.

So the problem is that my file permissions are wrong on the script.  In the pom, you'll see 660 is applied to the files in this directory.  When I manually chmod to 770, it works w/o the /bin/bash.

Should I bother to fix this properly? (find a better way to handle the file attributes, or put a chmod in the post-install or something)

PR #85, opened 2018-04-11
Apf 997 create docker images for all connectors

SudharsanThumattiSathyamoorthy:
Added docker images for all connectors.

All docker images will be pushed to ```hub.docker.com``` (```ws1connectors``` group) as a part of ```deploy``` phase.


pom.xml
```diff
@@ -441,11 +496,12 @@
                 <plugin>
                     <groupId>io.fabric8</groupId>
                     <artifactId>docker-maven-plugin</artifactId>
-                    <version>0.23.0</version>
+                    <version>0.25.0</version>
```
SudharsanThumattiSathyamoorthy: Updated the docker version to ```0.25.0```.

pom.xml
```diff
@@ -602,6 +657,107 @@
...
+                                        </assembly>
+                                        <from>openjdk:8u151-jre</from>
+                                        <workdir>/opt/vmware/${connector.name}</workdir>
+                                        <entryPoint>
+                                            <shell>
+                                                java -server -XX:+UnlockExperimentalVMOptions \
+                                                 -Xss256k \
+                                                 -XX:+UseCGroupMemoryLimitForHeap \
+                                                 -XX:MaxRAMFraction=1 \
+                                                 -XX:+HeapDumpOnOutOfMemoryError \
```
SudharsanThumattiSathyamoorthy: Added ```-XX:+HeapDumpOnOutOfMemoryError``` and ```-XX:HeapDumpPath=/tmp``` jvm options.

pom.xml
```diff
@@ -602,6 +657,107 @@
                 </plugins>
             </build>
         </profile>
+
```
SudharsanThumattiSathyamoorthy: Tested this profile on my local VM.

SudharsanThumattiSathyamoorthy: workspace one connectors

SudharsanThumattiSathyamoorthy: bitbucket server connector

connectors/airwatch/pom.xml
```diff
@@ -160,6 +161,87 @@
...
+                                        <assembly>
+                                            <inline>
+                                                <files>
+                                                    <file>
+                                                        <source>${project.basedir}/src/main/resources/application.yml</source>
+                                                        <outputDirectory>./</outputDirectory>
+                                                    </file>
+                                                    <file>
+                                                        <source>${project.basedir}/src/test/resources/docker/managed-apps.yml</source>
+                                                        <outputDirectory>./</outputDirectory>
```
rworsnop: Please keep in mind that these images must be usable by anyone. We should not be hard-wiring configuration like this.

The piece you are missing is probably that Docker lets you do this:
```
docker run <image> options...
```

So users should be able to do things like:
```
docker run -d -p 8080 ws1connectors/jira-connector --security.oauth2.resource.jwt.key-uri=http://foo
```
This would run the entrypoint, with all the args you've already specified, and then append the user's args. This doesn't seem to work at the moment, though. You might need to switch to `<exec>` instead of `<shell>`, but that's just a guess on my part.
Where the users wants to specify an entire configuration file, they can use volumes or bind mounts to allow the container to access the host file system.

connectors/airwatch/pom.xml
```diff
@@ -160,6 +161,87 @@
...
+                                                </files>
+                                            </inline>
+                                            <targetDir>/opt/vmware/airwatch-connector</targetDir>
+                                        </assembly>
+                                        <from>openjdk:8u151-jre</from>
+                                        <workdir>/opt/vmware/airwatch-connector</workdir>
+                                        <entryPoint>
+                                            <shell>
+                                                java -server -XX:+UnlockExperimentalVMOptions \
+                                                 -Xss256k \
```
rworsnop: We don't need to change the stack size for connectors. This is useful for Zuul, which is still on the thread-per-request model.

connectors/airwatch/pom.xml
```diff
@@ -160,6 +161,87 @@
...
+                                                 -XX:+UseCGroupMemoryLimitForHeap \
+                                                 -XX:MaxRAMFraction=1 \
+                                                 -XX:+HeapDumpOnOutOfMemoryError \
+                                                 -XX:HeapDumpPath=/tmp \
+                                                 -jar /opt/vmware/airwatch-connector/airwatch-connector-${project.version}.jar \
+                                                 --logging.level.com.vmware=${container.logging.level} \
+                                                 --spring.profiles.active=prod,external \
```
rworsnop: The `external` profile is obsolete.

connectors/airwatch/pom.xml
```diff
@@ -160,6 +161,87 @@
...
+                                                 --tenant.publicKeyUrl="https://herocard.vmwareidentity.com/SAAS/API/1.0/REST/auth/token?attribute=publicKey&amp;format=pem" \
```
rworsnop: `tenant.publicKeyUrl` is obsolete.

connectors/airwatch/pom.xml
```diff
@@ -160,6 +161,87 @@
...
+                                                 --security.oauth2.resource.jwt.key-uri="https://herocard.vmwareidentity.com/SAAS/API/1.0/REST/auth/token?attribute=publicKey&amp;format=pem" \
```
rworsnop: We should not be hardwiring this.

connectors/airwatch/pom.xml
```diff
@@ -160,6 +161,87 @@
...
+                                                 --spring.config.location=/opt/vmware/airwatch-connector/managed-apps.yml,/opt/vmware/airwatch-connector/application.properties  \
```
rworsnop: There's no point burying config in the container. The user needs control over this.

connectors/airwatch/pom.xml
```diff
@@ -160,6 +161,87 @@
...
+                                                 --hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds=10000
```
rworsnop: Connectors don't use Hystrix.

connectors/airwatch/pom.xml
```diff
@@ -160,6 +161,87 @@
...
+                                            </shell>
+                                        </entryPoint>
+                                    </build>
+                                    <run>
```
rworsnop: The user runs the container; we don't. Therefore we don't need the `<run>` element at all.

connectors/airwatch/pom.xml
```diff
@@ -160,6 +161,87 @@
...
+                                                        <source>${project.basedir}/src/test/resources/docker/application.properties</source>
+                                                        <outputDirectory>./</outputDirectory>
+                                                    </file>
+
+                                                </files>
+                                            </inline>
+                                            <targetDir>/opt/vmware/airwatch-connector</targetDir>
+                                        </assembly>
+                                        <from>openjdk:8u151-jre</from>
+                                        <workdir>/opt/vmware/airwatch-connector</workdir>
```
rworsnop: It would be good to add a `<healthCheck>` element:
```
<healthCheck>
    <interval>20s</interval>
    <timeout>5s</timeout>
    <retries>6</retries>
    <cmd>curl -f http://localhost:8080/health || exit 1</cmd>
</healthCheck>
```

This won't actually do anything except display health in `docker ps` and `docker inspect`, but it gives users options to do things with that information.

connectors/airwatch/pom.xml
```diff
@@ -160,6 +161,87 @@
...
+                                                    <file>
+                                                        <source>${project.basedir}/src/test/resources/docker/application.properties</source>
+                                                        <outputDirectory>./</outputDirectory>
+                                                    </file>
+
+                                                </files>
+                                            </inline>
+                                            <targetDir>/opt/vmware/airwatch-connector</targetDir>
+                                        </assembly>
+                                        <from>openjdk:8u151-jre</from>
```
rworsnop: Wrong base image.

pom.xml
```diff
@@ -18,6 +18,11 @@
 
     <properties>
         <connector.name>overrideme</connector.name>
+        <container.memory>512000000</container.memory>
```
rworsnop: Don't need this.

pom.xml
```diff
@@ -18,6 +18,11 @@
...
+        <container.logging.level>DEBUG</container.logging.level>
+        <docker.javaImage>ws1devops/java8:u162</docker.javaImage>
+        <docker.wait.timeout>300000</docker.wait.timeout>
```
rworsnop: Don't need this.

pom.xml
```diff
@@ -18,6 +18,11 @@
...
+        <docker.bind.address>172.17.0.1</docker.bind.address>
```
rworsnop: Don't need this.

pom.xml
```diff
@@ -602,6 +657,107 @@
...
+                                                        <useProjectArtifact>true</useProjectArtifact>
+                                                        <includes>
+                                                            <include>com.vmware.card-connectors:${project.artifactId}</include>
+                                                        </includes>
+                                                    </dependencySet>
+                                                </dependencySets>
+                                            </inline>
+                                            <targetDir>/opt/vmware/${connector.name}</targetDir>
+                                        </assembly>
+                                        <from>openjdk:8u151-jre</from>
```
rworsnop: My comments on the airwatch connector POM also apply here.

pom.xml
```diff
@@ -602,6 +657,107 @@
...
+                                </image>
+                            </images>
+                        </configuration>
+                        <executions>
+                            <execution>
+                                <id>build</id>
+                                <phase>package</phase>
+                                <goals>
+                                    <goal>build</goal>
+                                    <goal>start</goal>
```
rworsnop: We don't need `start` and `stop`.

pom.xml
```diff
@@ -602,6 +657,107 @@
...
+                <plugins>
+                    <plugin>
+                        <groupId>io.fabric8</groupId>
+                        <artifactId>docker-maven-plugin</artifactId>
+                        <version>0.25.0</version>
+
+                        <configuration>
+                            <registry>registry.hub.docker.com/ws1connectors</registry>
+                            <authConfig>
+                                <username>appplatform</username>
```
rworsnop: Please parameterize these credentials and we can set them in Bamboo. Credentials don't belong in GitHub.

connectors/airwatch/pom.xml
```diff
@@ -160,6 +161,87 @@
...
+                                        <entryPoint>
+                                            <shell>
+                                                java -server -XX:+UnlockExperimentalVMOptions \
+                                                 -Xss256k \
+                                                 -XX:+UseCGroupMemoryLimitForHeap \
+                                                 -XX:MaxRAMFraction=1 \
+                                                 -XX:+HeapDumpOnOutOfMemoryError \
+                                                 -XX:HeapDumpPath=/tmp \
+                                                 -jar /opt/vmware/airwatch-connector/airwatch-connector-${project.version}.jar \
+                                                 --logging.level.com.vmware=${container.logging.level} \
```
RyanBard: Take this one out too.  Let the user define the logging.level of things.

pom.xml
```diff
@@ -18,6 +18,11 @@
 
     <properties>
         <connector.name>overrideme</connector.name>
+        <container.memory>512000000</container.memory>
+        <container.logging.level>DEBUG</container.logging.level>
```
RyanBard: I think this goes away when you remove the --logging.level

pom.xml
```diff
@@ -117,8 +122,58 @@
                 <artifactId>json-schema-validator</artifactId>
                 <version>2.2.6</version>
             </dependency>
-        </dependencies>
+            <dependency>
```
RyanBard: I don't see any new references to any of these dependencies.  What needed them?  Are they really needed in dependency management?

pom.xml
rworsnop: You shouldn't need to put all this at the root. `connectors/pom.xml` should work.

connectors/airwatch/pom.xml
SudharsanThumattiSathyamoorthy: Removed ```-Xss256k```.

connectors/airwatch/pom.xml
SudharsanThumattiSathyamoorthy: Removed ```external``` profile.

connectors/airwatch/pom.xml
SudharsanThumattiSathyamoorthy: Removed ```tenant.publicKeyUrl``` jvm option.

connectors/airwatch/pom.xml
SudharsanThumattiSathyamoorthy: Removed ```--security.oauth2.resource.jwt.key-uri``` jvm option.

connectors/airwatch/pom.xml
SudharsanThumattiSathyamoorthy: Removed all the configs.

connectors/airwatch/pom.xml
SudharsanThumattiSathyamoorthy: Removed ```Hystrix``` option.

connectors/airwatch/pom.xml
SudharsanThumattiSathyamoorthy: Removed ```<run>``` element.

connectors/airwatch/pom.xml
SudharsanThumattiSathyamoorthy: Added ```healthCheck``` element.

connectors/airwatch/pom.xml
SudharsanThumattiSathyamoorthy: Changed the base image to ```ws1devops/java8:u162```.

pom.xml
SudharsanThumattiSathyamoorthy: Removed ```container.memory```.

pom.xml
SudharsanThumattiSathyamoorthy: Removed ```docker.wait.timeout```.

pom.xml
SudharsanThumattiSathyamoorthy: Removed ```docker.bind.address```.

pom.xml
SudharsanThumattiSathyamoorthy: Done.

pom.xml
SudharsanThumattiSathyamoorthy: Removed ```start``` and ```stop``` goals.

pom.xml
SudharsanThumattiSathyamoorthy: Added ```${docker.hub.username}``` and ```${docker.hub.password}``` parameters for docker hub username and key.

connectors/airwatch/pom.xml
SudharsanThumattiSathyamoorthy: yes.. not required.. i have removed it.

pom.xml
SudharsanThumattiSathyamoorthy: yes.. not required.. i have removed it.

pom.xml
SudharsanThumattiSathyamoorthy: It was required, otherwise the docker build was failing. Now it is not required since we have moved the contents to ```connectors``` ```pom.xml```.

pom.xml
SudharsanThumattiSathyamoorthy: Moved the contents to ```connectors``` ```pom.xml```.

connectors/airwatch/pom.xml
SudharsanThumattiSathyamoorthy: Yes.. you are right... I have changed ```shell``` to ```exec```. Now we will be able to specify external configs.

```
docker run -it --mount type=bind,source="/Users/ssathiamoort/projects/github/connectors-workspace-one/connectors/airwatch/src/test/resources/rpmcheck",target=/opt/vmware/airwatch-connector ws1connectors/airwatch-connector /bin/sh --spring.config.location=file:/opt/vmware/airwatch-connector/managed-apps.yml --security.oauth2.resource.jwt.key-uri="https://herocard.vmwareidentity.com/SAAS/API/1.0/REST/auth/token?attribute=publicKey&format=pem" --greenbox.url=https://herocard.vmwareidentity.com
```

PR #90, opened 2018-04-19
Apf 1079 add mutually exclusive set id property to card action

SudharsanThumattiSathyamoorthy:


common/core/src/test/resources/simple.json
```diff
@@ -40,6 +40,7 @@
       }
     ],
     "allow_repeated":true,
-    "completed_label": "Completed"
+    "completed_label": "Completed",
+    "mutually_exclusive_set_id": "HCA_Approve_ServiceNow_1"
```
RyanBard: This test case isn't very meaningful (this property doesn't make sense if there is only 1 action with it).

Instead of this, go ahead and fix Concur connector to use this (and get rid of the Open button like Ashley recommended).

connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
```diff
@@ -268,15 +270,4 @@ private CardBodyField makeCardBodyField(final String title, final String descrip
                                 .build()
                 );
     }
-
-    private CardAction.Builder getOpenActionBuilder(final String baseUrl, final Locale locale) {
```
SudharsanThumattiSathyamoorthy: Per Ashley suggestion, removed ```Open``` action from concur connector.

RyanBard:
@rworsnop if you want it changed, probably speak your peace on APF-1039

PR #91, opened 2018-04-20
Apf 1014 avoid npe for unassigned mdm apps

shsmysore:
- NPE 1 : 
If app install card-action is performed for an app, which is added as MDM but not yet assigned.
Since the assignment is not done, this app will be missing in vIDM catalog.
( DS says the app's install status is `false` which leads to AirWatch card generation. )

- NPE 2 : 
There can be a missing configuration for any platform in `managed-apps.yml` file. Card request for this platform is now just skipped silently.

### DIFF
--- a/connectors/airwatch/src/test/resources/connector/requests/request.json
+++ b/connectors/airwatch/src/test/resources/connector/requests/request.json
@@ -2,6 +2,6 @@
   "tokens": {
     "udid": ["ABCD"],
     "platform": ["android"],
-    "app_keywords": ["Boxer", "Concur"]
+    "app_keywords": ["Boxer", "Concur", "browser"]
   }
 }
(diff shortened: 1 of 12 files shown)


connectors/airwatch/src/test/resources/connector/requests/request.json
```diff
@@ -2,6 +2,6 @@
   "tokens": {
     "udid": ["ABCD"],
     "platform": ["android"],
-    "app_keywords": ["Boxer", "Concur"]
+    "app_keywords": ["Boxer", "Concur", "browser"]
```
shsmysore: `browser` app is not configured for `android` platform.
Connector should just continue silently.

RyanBard: oh ok, cool, thanks

PR #92, opened 2018-04-28
APF-1021 Switch to WebTestClient/OkHttp mock

rworsnop:
Switched tests to WebTestClient for sending requests to the
connector and OkHttp for mocking the backends.

Using OkHttp means the connector is talking to the backend over HTTP just like in
production. We also have full-blown Tomcat starting up now, instead of mocked
servlets, so that’s more realistic as well.

For now, the AWS cert connector has been left out due to the complications around
it validating the backend host (which is always localhost when we’re using OkHttp.
There is a separate JIRA issue for dealing with this.


rworsnop:
Closing due to conflicts.

PR #93, opened 2018-04-29
APF-1021 Switch to WebTestClient/OkHttp mock

rworsnop:
Switched tests to WebTestClient for sending requests to the
connector and OkHttp for mocking the backends.

Using OkHttp means the connector is talking to the backend over HTTP just like in
production. We also have full-blown Tomcat starting up now, instead of mocked
servlets, so that’s more realistic as well.

For now, the AWS cert connector has been left out due to the complications around
it validating the backend host (which is always localhost when we’re using OkHttp.
There is a separate JIRA issue for dealing with this.


connectors/servicenow/src/test/java/com/vmware/connectors/servicenow/ServiceNowControllerTest.java
```diff
@@ -184,39 +163,38 @@ private ResultActions reject(String authToken, String requestFile) throws Except
...
+        String body = requestCards(SNOW_AUTH_TOKEN, "valid/cards/card.json")
+                .expectStatus().isOk()
+                .expectHeader().contentTypeCompatibleWith(APPLICATION_JSON)
+                .returnResult(String.class)
+                .getResponseBody()
+                .collect(Collectors.joining())
+                .map(JsonNormalizer::forCards)
+                .block();
+        assertThat(body, sameJSONAs(fromFile(
+                "/servicenow/responses/success/cards/email-not-found.json")).allowingAnyArrayOrdering());
```
RyanBard: email-not-found, empty-email, empty-tickets, missing-email, and missing-tickets are all empty card responses.  Does this expect get simpler if you remove the parts that prep for the allowingAnyArrayOrdering?

connectors/servicenow/src/test/resources/servicenow/responses/success/cards/card_xx.json
```diff
@@ -10,32 +10,32 @@
       "header": {
         "title": "[SERVICE NOW] APPROVAL REQUEST",
         "subtitle": [
-          "REQ0010002"
+          "REQ0010003"
```
RyanBard: Is the allowingAnyArrayOrdering not working?

rworsnop:
For some background, see the discussion here: https://jira.spring.io/browse/SPR-15286

Looks like I've added around 5 seconds to each connector. As far as I can tell, the OkHttp mock adds negligible overhead, which would suggest it's `WebEnvironment.RANDOM_PORT` that's the culprit. At some point in the future we should be able to use  `WebEnvironment.MOCK`. But according to [this documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-testing.html), "using WebTestClient in a mocked web application only works with WebFlux at the moment".  
Not really sure what you mean by adding things we can't control - probably best to discuss in person.

connectors/airwatch/src/test/java/com/vmware/connectors/airwatch/AirWatchControllerTest.java
```diff
@@ -157,32 +172,38 @@ void testRequestCardsSuccessiOS() throws Exception {
...
-                .andExpect(status().isOk());
+                .header("x-airwatch-base-url", mockBackend.url(""))
+                .body(BodyInserters.fromFormData("app_name", "Concur")
+                        .with("udid", "ABCD")
+                        .with("platform", "android"))
+                .exchange()
+                .expectStatus().isOk();
 
         mockBackend.verify();
+        mockGreenbox.verify();
```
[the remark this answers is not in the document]
rworsnop: Done. I've moved the `verify` call up to the base class.

connectors/gitlab-pr/src/main/java/com/vmware/connectors/gitlab/pr/GitlabPrController.java
```diff
@@ -189,7 +189,7 @@ private URI makeGitlabUri(
          */
         try {
             return new URI(
-                    baseUrl + "/api/v4/projects/" + mergeRequestId.getProjectId()
+                    baseUrl.replaceAll("/$", "") + "/api/v4/projects/" + mergeRequestId.getProjectId()
```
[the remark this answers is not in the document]
rworsnop: It's a regular expression. It's stripping the trailing "/" if it exists. I've added a comment explaining this.

connectors/servicenow/src/test/java/com/vmware/connectors/servicenow/ServiceNowControllerTest.java
rworsnop: Yep.

connectors/servicenow/src/test/resources/servicenow/responses/success/cards/card_xx.json
rworsnop: An oversight. Corrected.

PR #94, opened 2018-04-30
APF-1021 Allow zero or one calls that could fail in zip

rworsnop:


connectors/bitbucket-server/src/test/java/com/vmware/connectors/bitbucket/server/BitbucketServerControllerTest.java
```diff
@@ -139,11 +142,11 @@ private void buildRequestForCards() {
 
         expect(pr236Url).andRespond(withSuccess(pr236, APPLICATION_JSON));
         expect(pr246Url).andRespond(withSuccess(pr246, APPLICATION_JSON));
-        expect(notFoundUrl).andRespond(withStatus(HttpStatus.NOT_FOUND));
+        expect(between(0, 1), notFoundUrl).andRespond(withStatus(HttpStatus.NOT_FOUND));
```
RyanBard: What changed about this?  Why aren't we certain how many times the not-found-url is called?  Was the previous `expect.andRespond` more of a `when.thenRespond` (I can't remember which, but I know I've had some libraries treat `expect` as a mock/verification and `when` as a stub)?

rworsnop: `.zip` won't bother with any other `Mono`s if one fails.
So if we get the 404 from "/" before we get chance to hit "/activities", `verify` will fail.
I'm not  sure how it was working before. If anything I would expect the second call to be missed all the time because the first one should "respond" immediately with 404 every time.
Another approach I considered was using `zipDelayError` instead of `zip`. This would guarantee that both calls were made, regardless of whether or not one failed. I thought it was better to change the tests to match behavior instead of changing behavior to match the tests, though.

RyanBard: There might be a timing or order difference now that they're going through a real http server.  Do you know if OkHttp serves requests in parallel?  Also, was this now intermittently failing?

rworsnop: Yes and yes.

PR #101, opened 2018-05-08
Apf 1124 Add APIs for updating close date and next step

SudharsanThumattiSathyamoorthy:


SudharsanThumattiSathyamoorthy:
Tested the endpoints using postman. Will do end-end testing once APF-1125 & APF- 1156 changes are in.

connectors/salesforce/src/main/java/com/vmware/connectors/salesforce/SalesforceController.java
```diff
@@ -737,4 +741,48 @@ private String formatRecipient(UserRecord userRecord) {
...
+
+        return updateOpportunityField(baseUrl, auth, opportunityId, body);
+    }
+
+    private Mono<Void> updateOpportunityField(final String baseUrl,
+                                              final String auth,
+                                              final String opportunityId,
+                                              final Object body) {
+        return rest.patch()
+                .uri(baseUrl + "/services/data/v39.0/sobjects/Opportunity/{opportunityId}", opportunityId)
```
shsmysore: We are using v20.0 for all the existing functions.
Does any of these actions needs to bump the version ?

https://github.com/vmware/connectors-workspace-one/blob/master/connectors/salesforce/src/main/resources/application.yml

SudharsanThumattiSathyamoorthy: When the existing functionalities were built,```v20.0``` would have been the latest. Currently, ```v39.0``` is the API latest version. It is always better to go for latest API version.

RyanBard: +1 for updating all of the versions to the latest (be sure to manually test that it all still works)

Also, to keep everything in the same place and able to be overridden (either everything should be overridable or none of it should be imo), move the url into the yml like the others.

SudharsanThumattiSathyamoorthy: I have moved the API to ```application.yml``` file. I have to go through the salesforce API document to see if there are any side-effects of upgrading the existing APIs to the latest version.

I have created a follow-up ticket APF-1159 to update the existing APIs to ```v39.0```.

connectors/salesforce/src/main/resources/application.yml
```diff
@@ -10,6 +10,7 @@ sf:
    opportunityContactLinkPath: services/data/v20.0/sobjects/OpportunityContactRole
    opportunityTaskLinkPath: services/data/v20.0/sobjects/Task
    attachmentTasklinkPath: services/data/v20.0/sobjects/Attachment
+   opportunityFieldsUpdatePath: services/data/v39.0/sobjects/Opportunity/{opportunityId}
```
shsmysore: Should we need to update all of them to latest v39.0 ?

RyanBard: He created a followup ticket for it: https://jira-euc.eng.vmware.com/jira/browse/APF-1159

SudharsanThumattiSathyamoorthy: @shsmysore : Yeah, we should upgrade all the API's to ```v39.0``` if it does not hamper any existing functionality.

PR #104, opened 2018-05-16
APF-1170 Validations for managed-app configurations

shsmysore:
- Check if managed-apps.yml configuration is missing.
- greenbox.url
- App name or Id may be missing for apps on each platform.

managed-apps

[error-logs.txt](https://github.com/vmware/connectors-workspace-one/files/2010479/error-logs.txt)

### DIFF
--- a/connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/ConnConfiguration.java
+++ b/connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/ConnConfiguration.java
@@ -6,8 +6,11 @@
 package com.vmware.connectors.airwatch;
 
 import com.vmware.connectors.airwatch.config.AppConfigurations;
+import com.vmware.connectors.airwatch.exceptions.AppConfigException;
 import com.vmware.connectors.airwatch.service.AppConfigService;
 import org.apache.commons.io.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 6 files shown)


connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/ConnConfiguration.java
```diff
@@ -65,7 +84,11 @@ public String connectorMetadata() throws IOException {
 
     @Bean
     public URI gbBaseUrl() {
-        String gbUrl = environment.getProperty("greenbox.url");
-        return fromHttpUrl(gbUrl).build().toUri();
+        try {
+            return fromHttpUrl(environment.getProperty("greenbox.url")).build().toUri();
+        } catch (IllegalArgumentException ex) {
+            logger.error("Greenbox URL config is invalid. ", ex);
+            return null;
```
rworsnop: Would it be better if this method threw an exception instead of returning null? My concern is that returning null will just cause confusing errors later on. I haven't tried it though.

shsmysore: Returning null makes the application fail to start, since it can't find the required bean.
I'm attaching logs to quickly see how it looks.

[gb-url-missing-logs.txt](https://github.com/vmware/connectors-workspace-one/files/2013547/gb-url-missing-logs.txt)

shsmysore: Hmm.., this seems better. I'll fix it.
Then we can see custom exception and also bean creation exception in the logs.

PR #123, opened 2018-07-30
APF-1249: Fix SOQL injection

RyanBard:
* Fixes SOQL injection by escaping backslash and single quotes


RyanBard:
The SOQL injection before this PR wasn't successful b/c the first query could successfully get all users in that table, however, the followup query for opportunities happened to be invalid from Salesforce's point of view (some complaint about the AND or OR or something).

This fixes it so it correctly encodes the input, causing it not to potentially fall through in unexpected ways, and just returns empty cards results b/c the garbage email wasn't found.

I tried to write unit tests for this, but after a day and a half of fighting the very fragile salesforce connector unit tests, I just created another ticket to deal with that later: APF-1396

PR #125, opened 2018-07-31
APF-1356 Connector api should support object types

shsmysore:
To merger after https://jira-euc.eng.vmware.com/jira/browse/APF-1432

- Updated metadata JSON structure to the [new design.](https://confluence-euc.eng.vmware.com/display/AP/Engineering+Design%3A+Supporting+Multiple+Object+Types+in+a+Single+Connector)

- Standard set of X-headers.
(https://jira-euc.eng.vmware.com/jira/browse/APF-1357)

- Cache control header for metadata JSON.

- Removed discovery HAL resource.

- Serve metadata JSON directly from the root.
This would be simple and easy for admins to configure.


connectors/airwatch/src/test/java/com/vmware/connectors/airwatch/AirWatchControllerTest.java
```diff
@@ -107,12 +107,20 @@ void testDiscovery() throws IOException {
...
         webClient.get()
                 .uri("/discovery/metadata.json")
                 .headers(ControllerTestsBase::headers)
                 .exchange()
                 .expectStatus().is2xxSuccessful()
-                .expectBody().json(fromFile("/connector/responses/metadata.json"));
+                .expectBody()
+                .consumeWith(res -> {
+                    String body = new String(res.getResponseBody());
+                    assertThat(body.replaceAll(localHostRegex, "CONNECTOR_HOST"),
```
SudharsanThumattiSathyamoorthy: Instead of ```CONNECTOR_HOST```, we can use ```http://localhost```.

SudharsanThumattiSathyamoorthy: The objective of the change is to replace ```CONNECTOR_HOST``` with the server URL in an actual scenario. It will be confusing when we change ```CONNECTOR_HOST``` again with ```CONNECTOR_HOST```. It would be better if we change ```CONNECTOR_HOST``` with some static URL like ```http://localhost``` and check with the response.

shsmysore: Yes, readability can be improved in Airwatch connector UT. It is having a separate metadata JSON file for unit tests. 
For others, I still use the same static metadata file, which is used by the connector controller. (In common ControllerTestsBase)

Done.

connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/AirWatchController.java
```diff
@@ -89,9 +91,10 @@ public AirWatchController(WebClient rest, CardTextAccessor cardTextAccessor,
         this.gbBaseUri = gbBaseUri;
     }
 
-    @GetMapping(path = "/discovery/metadata.json")
-    public ResponseEntity<String> getmetadata() {
-        return ResponseEntity.ok(connectorMetadata);
+    @GetMapping(path = METADATA_PATH)
+    public ResponseEntity<String> getmetadata(HttpServletRequest request) {
+        String requestUrl = request.getRequestURL().toString();
+        return ResponseEntity.ok(connectorMetadata.replace("CONNECTOR_HOST", requestUrl.split(METADATA_PATH)[0]));
```
RyanBard: This isn't going to work.

You need to take into account:
* X-Forwarded-Host
* X-Forwarded-Port
* X-Forwarded-Proto

I'll dig up where this is happening in discovery.

Until then, you can test how this works by using curl (or postman) to interact with a connector's discovery:

```
curl -v \
     -H "Authorization: Bearer <vidm-token>" \
     -H "X-Forwarded-Proto: https" \
     -H "X-Forwarded-Host: www.my-host.com" \
     -H "X-Forwarded-Port: 8443" \
     "http://localhost:8080"
```

RyanBard: You have to make sure you use UriComponentsBuilder.fromHttpRequest(HttpRequest) like we do in CommonUtils: https://github.com/vmware/connectors-workspace-one/blob/7c6bb345131f8a3c1e64eb67333a4cda90d26ffd/common/connectors-config/src/main/java/com/vmware/connectors/common/utils/CommonUtils.java#L41

That spring code is what is respecting the X-Forwarded headers.

connectors/airwatch/src/main/resources/static/discovery/metadata.json
```diff
@@ -1,16 +1,28 @@
...
+        "platform": {
+          "env": "DEVICE_SYSTEM_NAME"
+        },
+        "app_keywords": {
+          "capture_group": 1,
+          "regex": "(CONNECTOR_REGEX)"
+        }
+      },
+      "endpoint": {
+        "href": "CONNECTOR_HOST/cards/requests"
```
RyanBard: This should end up `"https://my-connector/cards/requests"` because of the X-Forwarded headers set in ControllerTestsBase: https://github.com/vmware/connectors-workspace-one/blob/7c6bb345131f8a3c1e64eb67333a4cda90d26ffd/common/connectors-test/src/main/java/com/vmware/connectors/test/ControllerTestsBase.java#L144

connectors/aws-cert/src/main/java/com/vmware/connectors/aws/cert/AwsCertController.java
```diff
@@ -50,26 +54,37 @@
...
         this.certificateApprovalPath = certificateApprovalPath;
+        this.metadata = IOUtils.toString(metadataJsonResource.getInputStream(), Charset.defaultCharset());
         this.rest = rest;
         this.cardTextAccessor = cardTextAccessor;
     }
 
+    @GetMapping(path = METADATA_PATH)
+    public ResponseEntity<String> getMetadata(HttpServletRequest request) {
+        String requestUrl = request.getRequestURL().toString();
+        return ResponseEntity.ok(this.metadata.replace("CONNECTOR_HOST", requestUrl.split(METADATA_PATH)[0]));
```
RyanBard: If you're going to do a string replace on the metadata.json contents, use something less likely to come up in a real url and more obvious that it will be replaced: `@@CONNECTOR_HOST@@` or `${CONNECTOR_HOST}` or something

connectors/airwatch/src/test/java/com/vmware/connectors/airwatch/AirWatchControllerTest.java
```diff
@@ -107,12 +107,24 @@ void testDiscovery() throws IOException {
...
         webClient.get()
                 .uri("/discovery/metadata.json")
                 .headers(ControllerTestsBase::headers)
                 .exchange()
                 .expectStatus().is2xxSuccessful()
-                .expectBody().json(fromFile("/connector/responses/metadata.json"));
+                .expectBody()
+                .consumeWith(res -> {
+                    String body = new String(res.getResponseBody());
+                    assertThat(body.replaceAll(localHostRegex, "http://localhost"),
```
RyanBard: This goes back to a simple json file when you fix the X-Forwarded problem.

common/connectors-test/src/main/java/com/vmware/connectors/test/ControllerTestsBase.java
```diff
@@ -130,14 +125,23 @@ protected void testConnectorDiscovery() throws IOException {
...
                 .headers(ControllerTestsBase::headers)
                 .exchange()
                 .expectStatus().is2xxSuccessful()
-                .expectBody().json(fromFile("/static/discovery/metadata.json"));
-
-        verifyObjectTypeField();
+                .expectBody()
+                .consumeWith(res -> {
+                    String body = new String(res.getResponseBody());
+                    assertThat(body.replaceAll(localHostRegex, "CONNECTOR_HOST"),
```
RyanBard: Here too (don't need the string replaceAll after fixing the X-Forwarded issue).

connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/AirWatchController.java
shsmysore: Thanks, I didn't think about 'X-Forwarded' at first.
Fixed it and tested from Postman.

connectors/aws-cert/src/main/java/com/vmware/connectors/aws/cert/AwsCertController.java
shsmysore: Yes, also this might help quickly identify there is something to be replaced, while reading.

connectors/bitbucket-server/src/main/resources/static/discovery/metadata.json
```diff
@@ -1,11 +1,23 @@
 {
   "authorization_header": "x-bitbucket-server-authorization",
   "backend_base_url_header": "x-bitbucket-server-base-url",
```
rworsnop: Once I see that the new format is being used, I'm not going to read these values. I expect "new" connectors to look for 'X-Connector-Authorization` and `X-Connector-Base-Url`.

shsmysore: Cool, I'll add changes to this branch itself.

shsmysore: Connectors now look for standard set of headers.

connectors/bitbucket-server/src/main/resources/application.yml
```diff
@@ -4,9 +4,6 @@ server:
      mime-types: application/json,application/xml,text/html,text/xml,text/plain
   server-header: bitbucket-server-connector
 
-connector:
-  hasTestAuth: true
```
RyanBard: I probably would have forgotten about this :)

connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/AirWatchController.java
```diff
@@ -78,20 +83,29 @@
...
+        this.maxAge = maxAge;
+        this.unit = unit;
     }
 
-    @GetMapping(path = "/discovery/metadata.json")
-    public ResponseEntity<String> getmetadata() {
-        return ResponseEntity.ok(connectorMetadata);
+    @GetMapping(path = "/")
+    public ResponseEntity<String> getMetadata(HttpServletRequest request) {
+        return ResponseEntity.ok()
```
rworsnop: This is repeated for each connector. I would think that modifying the common root controller instead of removing it would be the way to go here.

shsmysore: This is good idea.
Done.

PR #127, opened 2018-08-03
APF-1403: Downgrade jjwt to 0.9.1

RyanBard:
* Downgrades jjwt from 0.10.1 to 0.9.1 until vIDM key strength is increased

### DIFF
--- a/common/core-test/pom.xml
+++ b/common/core-test/pom.xml
@@ -37,18 +37,6 @@
             <artifactId>spring-hateoas</artifactId>
         </dependency>
         <dependency>
-            <groupId>io.jsonwebtoken</groupId>
-            <artifactId>jjwt-api</artifactId>
-        </dependency>
-        <dependency>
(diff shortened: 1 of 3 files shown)


RyanBard:
Unfortunately, I had to drop this back down to 0.9.1 (so we went from 0.9.0 to 0.10.1 down to 0.9.1, so still an update overall) because the JWTs from our vIDM systems are signed w/ a 1024 bit key which is weak enough to throw exceptions in jjwt (and there doesn't appear to be a way to disable that check in the interim).

common/core-test/pom.xml
```diff
@@ -36,18 +36,6 @@
             <groupId>org.springframework.hateoas</groupId>
             <artifactId>spring-hateoas</artifactId>
         </dependency>
-        <dependency>
```
RyanBard: This wasn't actually used.

PR #129, opened 2018-08-10
Apf 1394 make concur fail safe for invalid report ids

SudharsanThumattiSathyamoorthy:
Tested the changes with a valid and invalid report id. I was able to get the card result for the valid report.

### DIFF
--- a/connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
+++ b/connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
@@ -20,6 +20,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.core.io.Resource;
 import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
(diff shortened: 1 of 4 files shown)


connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
```diff
@@ -199,7 +200,7 @@ private String getConcurRequestTemplate(final String reason,
 
         return fetchOAuthToken(authHeader, clientId, clientSecret)
                 .flatMap(oauthHeader -> getReportDetails(oauthHeader, id, baseUrl))
-                .onErrorResume(Reactive::skipOnNotFound)
+                .onErrorResume(throwable -> Reactive.skipOnStatus(throwable, HttpStatus.BAD_REQUEST))
```
SudharsanThumattiSathyamoorthy: In actual scenario, concur send a bad request ```400``` for invalid report ids.

connectors/concur/src/test/java/com/vmware/connectors/concur/ConcurControllerTest.java
```diff
@@ -152,6 +153,22 @@ void testRequestCardsSuccess(String lang, String resFile) throws Exception {
...
 
+    @Test
+    void testRequestCardsWithInvalidReportId() throws Exception {
+        expect(REPORT_ID_1).andRespond(withSuccess(
+                fromFile("/concur/responses/report_id_1.json")
+                        .replace("${concur_host}", mockBackend.url("")), APPLICATION_JSON));
+
+        expect(INVALID_REPORT_ID).andRespond(withStatus(HttpStatus.BAD_REQUEST));
+
+        mockConcurServer.expect(ExpectedCount.manyTimes(), requestTo("/oauth2/v0/token"))
```
RyanBard: Is this actually called multiple times?  I thought the token call is only once and the same token is used for multiple report calls.

SudharsanThumattiSathyamoorthy: The token call is made for every report. We can optimize by caching the token. I have created a jira ticket for the same.

https://jira-euc.eng.vmware.com/jira/browse/APF-1440

RyanBard: You don't even need to cache.  Just make the first call retrieve the oauth token and pass that along and do the other calls per cards token.

PR #131, opened 2018-08-14
Apf 1395 change concur regex to extract report

SudharsanThumattiSathiamoorthy:


connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
```diff
@@ -105,12 +107,30 @@ public ConcurController(WebClient rest,
             @Valid @RequestBody CardRequest cardRequest,
             final HttpServletRequest request) {
 
+        final String automatedEmailSubject = cardRequest.getTokenSingleValue(CONCUR_AUTOMATED_EMAIL_SUBJECT);
```
SudharsanThumattiSathyamoorthy: Added this check to avoid false positives.

connectors/concur/src/main/java/com/vmware/connectors/concur/ConcurController.java
```diff
@@ -105,12 +107,30 @@ public ConcurController(WebClient rest,
...
+        if (StringUtils.isBlank(automatedEmailSubject)) {
+            return Mono.just(new Cards());
+        }
+
         final Set<String> expenseReportIds = cardRequest.getTokens(EXPENSE_REPORT_ID);
+        if (CollectionUtils.isEmpty(expenseReportIds)) {
+            return Mono.just(new Cards());
+        }
 
+        return fetchOAuthToken(authHeader, clientId, clientSecret)
```
SudharsanThumattiSathyamoorthy: Modified the logic to fetch the OAuth token only once.

PR #135, opened 2018-08-21
APF-1357 Snapshot version updated to 2.3

shsmysore:
Since we switched to multi-object type, I thought we can make it 2.3 Snapshot.
Then it'll be easy to use for snapshots in other repos. (Like SFDC internal)


RyanBard:
We shouldn't have to do this.  @rworsnop is going to deploy updated 2.2-SNAPSHOT.  Let me (or Rob) know if it doesn't work out.

rworsnop:
I have deployed the latest 2.2 snapshot to Sonatype.

PR #142, opened 2018-09-24
Apf 1490 create hash field in card schema

SudharsanThumattiSathyamoorthy:
1. Create hash field and the value will now be populated for all the connectors.
2. I have excluded some fields which makes the 'hash' field unreliable.
3. I have optimized imports across the connectors to remove unused imports in test packages. So the diff will contain some files only with import changes.


connectors/airwatch/pom.xml
```diff
@@ -20,7 +20,6 @@
         <dependency>
             <groupId>com.sun.activation</groupId>
             <artifactId>javax.activation</artifactId>
-            <version>1.2.0</version>
```
SudharsanThumattiSathyamoorthy: Removed ```javax.activation``` version from all the child pom files.

common/core/src/main/java/com/vmware/connectors/common/payloads/response/CardAction.java
```diff
@@ -54,6 +54,7 @@
     private boolean removeCardOnCompletion;
 
     @JsonProperty("request")
+    @ToStringExclude
```
RyanBard: You can't exclude this.  This is important information for the action.  Sometimes it contains IDs for the business object.  Ex. airwatch's install action: https://github.com/vmware/connectors-workspace-one/blob/120cbca3f9f85d4b34044862e0ca28f5bca9c363/connectors/airwatch/src/main/java/com/vmware/connectors/airwatch/AirWatchController.java#L223

common/core/src/main/java/com/vmware/connectors/common/payloads/response/Card.java
```diff
@@ -360,9 +373,34 @@ public Builder addTag(String tag) {
...
         }
+
+        private String computeHash() {
+            String result = "";
+            if (card.name != null) {
+                result += card.name;
+            }
+
+            if (card.header != null) {
+                result += card.header.toString();
```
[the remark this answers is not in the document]
SudharsanThumattiSathyamoorthy: Calculating hash on ```Map``` and ```Set``` objects make ```hash``` value change every time. I have tried including these fields, but ```hash``` value tend to change every time.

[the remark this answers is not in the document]
SudharsanThumattiSathyamoorthy: I tried even using LinkedHashMap and SortedMap implementations. But ```hash``` values tends to change every time.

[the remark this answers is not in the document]
SudharsanThumattiSathyamoorthy: Even if we write ```computeHash``` method inside ```CardHeader```, ```CardAction```, ```CardBody```, finally we will end up converting each field in that class to ```ToString``` and calculating```SHA-1```. 

In this case, if we add any new fields we might miss to include adding the new field in ```hash``` calculation.

When we use ````ToStringBuilder.reflectionToString(this, SHORT_PREFIX_STYLE);```, the new fields are automatically included in the hash computation, since we are calling ```toString``` from the ```computeHash``` method.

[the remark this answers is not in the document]
RyanBard: If your library doesn't have a way to hash Set/Map independent of order, then you will have to sort it so it is consistent (sort the Set by its values, set the Map by its keys).

common/core/src/main/java/com/vmware/connectors/common/payloads/response/Card.java
```diff
@@ -360,9 +373,34 @@ public Builder addTag(String tag) {
...
+            if (!CollectionUtils.isEmpty(card.actions)) {
+                for (CardAction action: card.actions) {
+                    result += action.toString();
+                }
+            }
+
+            if (card.body != null) {
+                result += card.body.toString();
+            }
+
```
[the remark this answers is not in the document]
SudharsanThumattiSathyamoorthy: Having any``` Set``` and ```Map``` objects makes the ```hash``` value change every time. So I have excluded these fields intentionally.

common/core/src/main/java/com/vmware/connectors/common/payloads/response/CardAction.java
SudharsanThumattiSathyamoorthy: Having any``` Set``` and ```Map``` objects makes the ```hash``` value change every time. So I have excluded these fields intentionally.

common/core/src/main/java/com/vmware/connectors/common/payloads/response/Card.java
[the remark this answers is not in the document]
RyanBard: If it's not clear, what I'm saying is: You have to extract the values out of the set/map in a certain order when computing the hash.

You aren't going to be able to just toString the Set/Map (again, we shouldn't be using toString for this computation anyway).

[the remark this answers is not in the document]
SudharsanThumattiSathyamoorthy: Got it.. Will change the code accordingly..!!

connectors/aws-cert/src/test/java/com/vmware/connectors/aws/cert/AwsCertControllerTest.java
```diff
@@ -124,6 +124,7 @@ void testRequestCardsSuccess(String lang, String responseFile) throws Exception
                 .collect(Collectors.joining())
                 .map(JsonNormalizer::forCards)
                 .block();
+        body = body.replaceAll("[a-z0-9]{40,}", "test-hash");
```
SudharsanThumattiSathyamoorthy: The ```${backend_host}``` placeholder in ```hero_aws_cert_approval_url``` is replaced with mock provider host and port. So the hash value is getting changed every time. Because of this we cannot compare with actual hash value for ```AwsCertController```. I have to replace the actual hash value with ```test-hash```.

There is a similar issue with ```JiraConnector``` as well.

connectors/aws-cert/src/test/resources/awscert/responses/success/cards/single-card.json
```diff
@@ -79,7 +79,8 @@
       ],
       "image": {
         "href": "https://my-connector/images/connector.png"
-      }
+      },
```
SudharsanThumattiSathyamoorthy: Card action url in JiraConnector is substituted with mock provider host and port. So the hash value was getting changed every time when we run the test case. I am replacing the actual has h value with ```test-hash``` before comparing the json value.

SudharsanThumattiSathyamoorthy:
allow connector author to set the hash if they choose
[Yes, added a setter method in Card builder]

don't use toString to compute the hash
[I have changed the logic now. I am concatenating each field and computing the hash value]

add card.tags to the hash
don't exclude card action's url from the hash
don't exclude card action's request params from the hash
maybe don't exclude card user input's options from the hash
[Except for UUID and date fields, I have included all fields for hash calculation.]

common/core/src/main/java/com/vmware/connectors/common/payloads/response/Card.java
```diff
@@ -353,16 +367,68 @@ public Builder addTag(String tag) {
...
+                        .filter(Objects::nonNull)
+                        .forEach(cardAction -> result.append(cardAction.hash()));
+            }
+
+            if (card.image != null && StringUtils.isNotBlank(card.image.getHref())) {
+                result.append(card.image.getHref());
+            }
+
+            if (!CollectionUtils.isEmpty(card.tags)) {
+                final Set<String> sortedSet = new TreeSet<>(card.tags);
```
RyanBard: Instead of using TreeSet because it happens to be sorted, just use java 8:

```java
card.tags.stream()
        .sorted()
        .forEach(tag -> result.append(tag));
```

common/core/src/main/java/com/vmware/connectors/common/payloads/response/CardAction.java
```diff
@@ -403,4 +403,49 @@ public CardAction build() {
             return completedAction;
         }
     }
+
+    public String hash() {
+        final StringBuilder result = new StringBuilder();
+
+        result.append(primary);
+
+        if (StringUtils.isNotBlank(label)) {
```
RyanBard: All of these null/empty checks are going to be a problem.  Take an object that looks like this:

```javascript
{
    x: 'a',
    y: 'b',
    z: 'c'
}
```
`sha1('abc')`

```javascript
{
    x: 'a',
    y: null,
    z: 'bc'
}
```
is also `sha1('abc')`

RyanBard: Similar to my concerns about toString, string concat (alone) is going to have some issues.

Quick googling didn't turn up anything I liked for computing a hash on objects.

We may end up having to write something that calculates this more generically ourselves (we could do something with serialization, but that introduces its own headaches).

Something that would be nice would be something that takes in varargs like `Objects.hash` (but not that, b/c the details of that are not appropriate for this -- maybe we can leverage it, but we have to be careful) or something uses annotations like `HashCodeBuilder.reflectionHashCode` (again, hashCode is not what we want; a hard-coded hashCode is a valid hashCode and we clearly wouldn't want that).

common/core/src/main/java/com/vmware/connectors/common/payloads/response/Card.java
```diff
@@ -353,16 +367,68 @@ public Builder addTag(String tag) {
             return this;
         }
 
+        /**
+         * Add SHA-1 of card fields.
+         *
+         * @param hash SHA-1 of card header, body, and action fields.
+         * @return this Builder instance, for method chaining
+         */
+        public Builder addHash(String hash) {
```
RyanBard: Also, this should be `setHash`, not `addHash`.  The `add` methods are because they add to a collection (`addTag` adds to the tags).

common/core/src/main/java/com/vmware/connectors/common/payloads/response/CardAction.java
SudharsanThumattiSathyamoorthy: I looked at the implementations of ```Objects.hash``` and ```HashCodeBuilder.reflectionHashCode```.

```Objects.hash``` just calculates the hash code of each object in the array and sum it up.
```HashCodeBuilder.reflectionHashCode``` is similar to the above one but it uses reflection to access the member fields and calculate hash code.

I tried to google about generating sha-1 for objects in an efficient. I did not find any relevant ideas or example. So we have to come with our solution.

One solution is adding an empty space or some separator before appending with the next field. For example, 

{
    x: 'a',
    y: 'b',
    z: 'c'
}

Appending separator would result: ```sha1(a b c)```

{
    x: 'a',
    y: null,
    z: 'bc'
}
  
Appending separator would result: ```sha1(a bc)```

Now this solution would avoid creating same hash for different objects.

We cannot remove ```null`` check, since ```DigestUtils.sha1Hex(null)``` is going to throw ```NullPointerException```. 

Please let me know the idea of adding a separator after each field and I will change the code accordingly.

File: common/core/src/main/java/com/vmware/connectors/common/payloads/response/Card.java (hunk 7 above)
Comment (SudharsanThumattiSathyamoorthy) [reply common/core/src/main/java/com/vmware/connectors/common/payloads/response/Card.java]: Done.

**Comment by SudharsanThumattiSathyamoorthy** [comment]
use java 8 streams instead of TreeSet to order things
[Fixed]
fix problems with using string concat to calculate the hash
[As we discussed, I have changed the hashing logic having ```field names + field values```]

File: common/core/src/main/java/com/vmware/connectors/common/payloads/response/Card.java [hunk 10]
```diff
@@ -353,16 +368,63 @@ public Builder addTag(String tag) {
...
+            final String headerHash = card.header == null ? null : card.header.hash();
+            final String bodyHash = card.body == null ? null : card.body.hash();
+            return HashUtil.hash(
+                    "name: ", card.name,
+                    "template: ", templateUrl,
+                    "header: ", headerHash,
+                    "body: ", bodyHash,
+                    "actions: ", actionHashList.toString(),
+                    "image: ", imageUrl,
+                    "tags: ", tagList.toString()
```
RyanBard:
```
System.out.println(List.of("a", "b, c"));
System.out.println(List.of("a", "b", "c"));
```

```
[a, b, c]
[a, b, c]
```

common/core/src/main/java/com/vmware/connectors/common/payloads/response/CardAction.java
```diff
@@ -403,4 +402,40 @@ public CardAction build() {
...
         }
     }
+
+    public String hash() {
+        final String url = this.url == null ? null : this.url.getHref();
+
+        final List<String> requestList = new ArrayList<>();
+        if (!CollectionUtils.isEmpty(request)) {
+            final Map<String, String> sortedMap = new TreeMap<>(request);
+            requestList.add(sortedMap.toString());
```
RyanBard:
```
System.out.println(Map.of("abc", "123", "xyz", "456"));
System.out.println(Map.of("abc", "123, xyz=456"));
```

```
{abc=123, xyz=456}
{abc=123, xyz=456}
```

common/core/src/main/java/com/vmware/connectors/common/payloads/response/CardActionInputField.java
```diff
@@ -240,4 +243,26 @@ private boolean validate(CardActionInputField field) {
             return true;
         }
     }
+
+    public String hash() {
+        final List<String> optionList = new ArrayList<>();
+        if (!CollectionUtils.isEmpty(this.options)) {
+            final Map<String, String> sortedMap = new TreeMap<>(this.options);
+            optionList.add(sortedMap.toString());
```
RyanBard:
```
System.out.println(Map.of("abc", "123", "xyz", "456"));
System.out.println(Map.of("abc", "123, xyz=456"));
```

```
{abc=123, xyz=456}
{abc=123, xyz=456}
```

common/core/src/main/java/com/vmware/connectors/common/payloads/response/CardBodyField.java
```diff
@@ -163,4 +168,28 @@ public CardBodyField build() {
...
         }
     }
+
+    public String hash() {
+        final List<String> contentList = new ArrayList<>();
+        if (!CollectionUtils.isEmpty(content)) {
+            for (Map<String, String> item : content) {
+                if (!CollectionUtils.isEmpty(item)) {
+                    final Map<String, String> sortedMap = new TreeMap<>(item);
+                    contentList.add(sortedMap.toString());
```
RyanBard:
```
System.out.println(Map.of("abc", "123", "xyz", "456"));
System.out.println(Map.of("abc", "123, xyz=456"));
```

```
{abc=123, xyz=456}
{abc=123, xyz=456}
```

common/core/src/main/java/com/vmware/connectors/common/payloads/response/CardHeader.java
```diff
@@ -44,4 +48,15 @@ public String getTitle() {
...
+
+    @Override
+    public String toString() {
+        return ToStringBuilder.reflectionToString(this, SHORT_PREFIX_STYLE);
+    }
+
+    public String hash() {
+        return HashUtil.hash(
+                "title: ", this.title,
+                "subtitle: ", subtitle.toString());
```
RyanBard:
```
System.out.println(List.of("a", "b, c"));
System.out.println(List.of("a", "b", "c"));
```

```
[a, b, c]
[a, b, c]
```

common/core/src/main/java/com/vmware/connectors/common/utils/HashUtil.java
```diff
@@ -0,0 +1,20 @@
+package com.vmware.connectors.common.utils;
+
+import org.apache.commons.codec.digest.DigestUtils;
+
+public final class HashUtil {
+
+    public static String hash(String... args) {
```
RyanBard: Make this `Object... args` so callers aren't forced to turn the args into strings.

common/core/src/main/java/com/vmware/connectors/common/payloads/response/CardActionInputField.java
```diff
@@ -240,4 +243,26 @@ private boolean validate(CardActionInputField field) {
...
+        }
+
+        return HashUtil.hash(
+                "id: ", this.id,
+                "label: ", this.label,
+                "format: ", this.format,
+                "options: ", optionList.toString(),
+                "minLength: ", String.valueOf(this.minLength),
```
RyanBard: You shouldn't need to do this if you change the HashUtil to take in `Object... args`.

SudharsanThumattiSathyamoorthy:
fix List -- add a "public static String hashList(List)"
[Done]
fix Map -- add a "public static String hashMap(Map<String, String>)"
[Done]
remove TreeSets
[I have removed it in the earlier commit]
use "Object... args" instead of "String... args" on hash method?
[I have changed to arg type to ```Object... args```]

RyanBard:
It looks ok to my eyes, but eyes don't make very good runtimes.

I'm going to review it some more this weekend and add some unit tests that I've started.

SudharsanThumattiSathyamoorthy:
Merged 143 into this PR.

PR #147, opened 2018-10-10
APF-1230: Tweaks for testing

RyanBard:


common/core/src/main/java/com/vmware/connectors/common/utils/AuthUtil.java
```diff
@@ -0,0 +1,28 @@
...
+
+public final class AuthUtil {
+
+    private static final ObjectMapper mapper = new ObjectMapper();
+
+    private AuthUtil() {
+        // hide constructor
+    }
+
+    public static String extractUserEmail(String authorization) throws IOException {
```
RyanBard: I moved this to a common place because the other Hub connectors will likely need it too.

